Issue #50, opened 2022-06-26
GetCredentialsBy {issuer, subject, schema} API calls ignoring query parameters


frankhinek:
## Expected Behavior

If a matching credential has been created, based on the following comment:

https://github.com/TBD54566975/ssi-service/blob/03d5868c5b2fcb8297cd586fafac0edb8f7e9cee/pkg/server/router/credential.go#L116

an API call to one of the following:

* `/v1/credentials?issuer=did:key:abc123`
* `/v1/credentials?schema=https://test-schema.com`
* `/v1/credentials?subject=did:key:def456`

should return a JSON array with the matching credential data.

## Current Behavior

If a matching credential has been created, an API call to one of the following:

* `/v1/credentials?issuer=did:key:abc123`
* `/v1/credentials?schema=https://test-schema.com`
* `/v1/credentials?subject=did:key:def456`

returns an error:

```bash
{
    "error": "must use one of the following query parameters: issuer, subject, schema"
}
```

## Steps to Reproduce

1. Create a credential:
    
    ```bash
    curl -X PUT -d '
     {
         "Issuer": "did:abc:123",
         "Subject": "did:def:456",
         "Schema": "https://test-schema.com/name",
         "Data": {
             "firstName": "Jack",
             "lastName": "Dorsey"
         },
         "Expiry": "2022-12-31T05:00:00+00:00"
     }' http://localhost:8080/v1/credentials
    ```
1. Attempt to query by issuer, subject, or schema:
    
    ```bash
    curl -X GET "http://localhost:8080/v1/credentials?issuer=did:abc:123"
    curl -X GET "http://localhost:8080/v1/credentials?subject=did:def:456"
    curl -X GET "http://localhost:8080/v1/credentials?schema=https://test-schema.com/name"
    ```
1. Observe that the following error is returned:
    
    ```bash
    {"error":"must use one of the following query parameters: issuer, subject, schema"}
    ```

## Possible Solution

Assuming that the design intent was for the API user to use a URL query string parameter/value to retrieve credentials, one possible fix is to create a new utility function in [/pkg/server/framework/util.go](https://github.com/TBD54566975/ssi-service/blob/03d5868c5b2fcb8297cd586fafac0edb8f7e9cee/pkg/server/framework/util.go):

```go
// GetQueryValue is a utility to get a parameter value from the query string, nil if not found
func GetQueryValue(r *http.Request, param string) *string {
	value := r.URL.Query().Get(param)
	if value == "" {
[...]

PR #54, opened 2022-07-09
Fix GetCredentialsBy API

atsushii:
Addresses #50 

What I did

The GetCredentials API does not receive query parameters properly.
Tried @frankhinek suggested method to solve this bug.
And that works!

Also added sanitization to logs to avoid injection attacks, I followed this recommendation method
 (https://codeql.github.com/codeql-query-help/go/go-log-injection/)


decentralgabe (approved):
nice and effective, thank you!

Issue #31, opened 2022-04-19
Add Support for OpenTelemetry


decentralgabe:
https://opentelemetry.io/docs/instrumentation/go/getting-started/

decentralgabe:
Handled in https://github.com/TBD54566975/ssi-service/pull/60 by @atsushii

PR #60, opened 2022-07-18
Add Support for OpenTelemetry

atsushii:
Addresses #31 

What I did

Added support for OpenTelemetry (https://opentelemetry.io/docs/instrumentation/go/getting-started/)

Please check the browser data on Jaeger

step

1, mage run
2, send a request e.g. curl http://localhost:8080/health
3, open the Jaeger UI (http://localhost:16686/)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,4 +29,7 @@ jobs:
       run: mage build
 
     - name: Test
-      run: mage test
+      run: mage CITest
+
+    - name: Upload coverage to Codecov
+      uses: codecov/codecov-action@v3
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 There are many ways to be an open source contributor, and we're here to help you on your way! You may:
 
-* Propose ideas in our [discussion forums](https://github.com/TBD54566975/ssi-service/discussions)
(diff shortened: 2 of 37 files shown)


decentralgabe:
Thank you @atsushii for your contribution. I sent a number of requests to the service while jager was running, but wasn't able to see the traces come through. I'm not familiar with OpenTelemetry so won't be of much use to you here.

Before merging, I think it's valuable to see traces for all requests that come into the service.

In this UI, I see only the jager service show up:
Screen Shot 2022-07-19 at 2 53 49 PM

atsushii:
@decentralgabe 
Did you reload the browser after sending the request? (I missed telling it..)
After reloading the browser, you can see the service like below!

Screen Shot 2022-07-20 at 9 07 44 AM

decentralgabe:
@atsushii I've tried reloading the browser. I even tried using multiple browsers (Safari, Chrome) after sending requests. Still don't see the SSI service show up.

atsushii:
@decentralgabe 
Thank you for trying!

I've deleted containers, images, and volumes before starting a new container then still it works.
I am not really sure why you can not see the SSI service but can you plz try to rebuild the container after deleting all related images, containers, and volumes if it is possible.😅

decentralgabe:
Removed all containers and did a fresh checkout. Worked!

decentralgabe (approved):

Issue #73, opened 2022-07-28
adding `-race` flag fails tests 


andorsk:
Note: https://github.com/TBD54566975/ssi-service/blob/main/magefile.go#L67 does not use the `-race` flag and when `-race` is applied, allocation errors are seen. 

Is there a good reason that the race flag is not included in these tests? 

```
go test -race  ./...  -v

fatal error: checkptr: converted pointer straddles multiple allocations

goroutine 11 [running]:
runtime.throw({0x10554e405?, 0x1053dc478?})
	/usr/local/go/src/runtime/panic.go:992 +0x50 fp=0xc00061eda0 sp=0xc00061ed70 pc=0x104d38880
runtime.checkptrAlignment(0xc00061ee08?, 0x104d55e20?, 0xc00006eb18?)
	/usr/local/go/src/runtime/checkptr.go:26 +0x80 fp=0xc00061edc0 sp=0xc00061eda0 pc=0x104d09c20
github.com/boltdb/bolt.(*Bucket).write(0xc00061ef68)
	/Users/akmb2/go/pkg/mod/github.com/boltdb/bolt@v1.3.1/bucket.go:626 +0x1b4 fp=0xc00061ee50 sp=0xc00061edc0 pc=0x1053dc604
github.com/boltdb/bolt.(*Bucket).CreateBucket(0xc000537358, {0xc00006eb18, 0x7, 0x8})
	/Users/akmb2/go/pkg/mod/github.com/boltdb/bolt@v1.3.1/bucket.go:188 +0x260 fp=0xc00061f010 sp=0xc00061ee50 pc=0x1053d95c0
github.com/boltdb/bolt.(*Bucket).CreateBucketIfNotExists(0xc000537340?, {0xc00006eb18, 0x7, 0x8})
	/Users/akmb2/go/pkg/mod/github.com/boltdb/bolt@v1.3.1/bucket.go:206 +0x48 fp=0xc00061f060 sp=0xc00061f010 pc=0x1053d97e8
github.com/boltdb/bolt.(*Tx).CreateBucketIfNotExists(...)
	/Users/akmb2/go/pkg/mod/github.com/boltdb/bolt@v1.3.1/tx.go:115
github.com/tbd54566975/ssi-service/pkg/storage.(*BoltDB).Write.func1(0xc000537340)
[...]

PR #74, opened 2022-07-28
updated bolt dependency and fix issue #73

andorsk:
related to issue https://github.com/TBD54566975/ssi-service/issues/73. `boltdb` is not long maintained, and a race condition occurs on the old build. With this update, you can add `-race` condition to magefile.

Included in this PR is also an addition of adding a `-race` flag to the mage tests.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,6 +23,8 @@ require (
 	gopkg.in/go-playground/validator.v9 v9.31.0
 )
 
+require go.etcd.io/bbolt v1.3.6 // indirect
+
 require (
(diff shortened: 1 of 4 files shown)


decentralgabe:
ah, I was avoiding updating because I thought bolt was stable!
I did try to debug this race condition issue for a few hours the other day and couldn't figure it out.
If the new lib fixes it, it's a good reason to update 😄 
thanks @andorsk !

andorsk:
@decentralgabe gracias! 🙇

Issue #78, opened 2022-08-07
Document build dependencies for Linux


ALRubinger:
Document installation of `go` and `mage` for Linux.

decentralgabe:
Pointing to the corresponding Mage and go documentation should be sufficient.

ALRubinger:
+1 - linked PR does this.

decentralgabe:
Fixed https://github.com/TBD54566975/ssi-service/commit/e0e2f33d36fc21b2ecf3746eda3638f8b3281664

PR #79, opened 2022-08-07
Issue #78: Add Linux installation instructions for Go and Mage

ALRubinger:


decentralgabe (changes requested):
I actually don't think it's worth having multi-platform instructions for a dependency (Mage), since Mage's own documentation covers this. Perhaps this PR can be amended to remove the MacOS specific instructions and just point to the Mage documentation.

ALRubinger:
Yup I think that's all it does -- links to Mage and shows Linux folks that's what they should use. I ported it from the updates I made to the template repo.

decentralgabe:
> Yup I think that's all it does -- links to Mage and shows Linux folks that's what they should use. I ported it from the updates I made to the template repo.

I mean having "MacOS" or "Linux" sections is unnecessary in itself. The tools themselves are multi-platform, and have supporting documentation for specific platforms, so pointing to them without calling out each is sufficient.

ALRubinger:
Oh, could be! I can alter the Mage section here and in the template repo to be "Multi-platform" instead of "macOS"-specific.

decentralgabe:
Fixed in the latest commit https://github.com/TBD54566975/ssi-service/commit/e0e2f33d36fc21b2ecf3746eda3638f8b3281664

If there are specific OS commands worth adding, we can amend. If we're just pointing to mage/go docs - that's covered.

Issue #75, opened 2022-08-01
Return Bad Request HTTP 400 context to client in error messages


ALRubinger:
Let's say I make a bad request to add a new VC. The verbose `curl` output I get back is:

```
% curl -X PUT -d '
{
    "Issuer": "did:key:zQ3shgaCdiqDT3K4ikNkZzbY7FHS8XNDhrCotGA813eRTJQAX",
    "Subject": "did:key:zQ3shgaCdiqDT3K4ikNkZzbY7FHS8XNDhrCotGA813eRTJQAX",
    "Schema": "59722df0-3458-4aeb-a5b4-929ad6bded63",
    "Data": {
        "emailAddress": “someone@somewhere.com”
    },
    "Expiry": "2022-12-31T05:00:00+00:00"
}' localhost:8080/v1/credentials -v
*   Trying 127.0.0.1:8080...
* Connected to localhost (127.0.0.1) port 8080 (#0)
> PUT /v1/credentials HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.79.1
> Accept: */*
> Content-Length: 319
> Content-Type: application/x-www-form-urlencoded
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 400 Bad Request
< Content-Type: application/json
< Date: Mon, 01 Aug 2022 17:19:36 GMT
< Content-Length: 45
<
* Connection #0 to host localhost left intact
{"error":"invalid create credential request"}
```

Trimming the verbosity, it's simply the JSON payload:

```
{"error":"invalid create credential request"}
```

The server logs show what's to be corrected in the request, but this won't be visible to the client:

```
build-web-1         | {"error":"invalid character 'â' looking for beginning of value","level":"error","msg":"invalid create credential request","time":"2022-08-01T17:21:33Z"}
```

I'm also not certain the error message on the server is correct for this example , but I'll look into that more and open another issue if so. This issue is for percolating necessary information to the client in `400 Bad Request` errors so they can correct and retry.

decentralgabe:
I agree the http should be minimal and simple; the logs should give more verbose information about what went wrong, with a request ID that can correlate the two.

ALRubinger:
Yep, keeping in mind that the client may not be able to access the server in a real-world case -- so showing the context behind the error (and as a bonus, how to potentially correct it!) is a great practice.

shobitb:
Regarding the actual error 
[...]

PR #84, opened 2022-08-11
Add context to error responses returned to client

shobitb:
This change preserves error context when sending error responses back to the client. It does so by using `framework.NewRequestError(err, httpCode)`` instead of `framework.NewRequestErrorMsg(errMsg, httpCode)` in places where the err object is available.

This fixes https://github.com/TBD54566975/ssi-service/issues/75


decentralgabe (approved):
thanks!

PR #106, opened 2022-09-20
SIP-3: KeyAccess API

decentralgabe:


internal/keyaccess/dataintegrity.go
```diff
@@ -0,0 +1,72 @@
...
+		return nil, errors.New("kid cannot be empty")
+	}
+	if key == nil {
+		return nil, errors.New("key cannot be nil")
+	}
+	publicKeyJWK, privateKeyJWK, err := crypto.PrivateKeyToPrivateKeyJWK(key)
+	if err != nil {
+		return nil, errors.Wrapf(err, "could not convert private key to JWK: %s", kid)
+	}
+	signer, err := cryptosuite.NewJSONWebKeySigner(kid, *privateKeyJWK, cryptosuite.AssertionMethod)
```
nitro-neal, line 35: Would AssertionMethod every need to be dynamic?

internal/keyaccess/dataintegrity.go
```diff
@@ -0,0 +1,72 @@
...
+		return nil, errors.Wrapf(err, "could not create JWK verifier: %s", kid)
+	}
+	return &DataIntegrityKeyAccess{
+		JSONWebKeySigner:   *signer,
+		JSONWebKeyVerifier: *verifier,
+		CryptoSuite:        cryptosuite.GetJSONWebSignature2020Suite(),
+	}, nil
+}
+
+func (ka DataIntegrityKeyAccess) Sign(payload cryptosuite.Provable) ([]byte, error) {
```
nitro-neal, line 50: Eventually we could make a class wrapper around []bytes throughout our codebase like a SignedEntity {signedJson []bytes}

Random []bytes could be error prone

internal/keyaccess/jwt.go
```diff
@@ -0,0 +1,85 @@
...
+	gocrypto "crypto"
+
+	"github.com/TBD54566975/ssi-sdk/credential"
+	"github.com/TBD54566975/ssi-sdk/credential/signing"
+	"github.com/TBD54566975/ssi-sdk/crypto"
+	"github.com/lestrrat-go/jwx/jwk"
+	"github.com/pkg/errors"
+)
+
+type JWKKeyAccess struct {
```
nitro-neal, line 13: Love these wrapper objects

internal/keyaccess/dataintegrity.go
decentralgabe: it does need to be dynamic, this would require some refactoring on the SDK side which I intend to cover in OSE-15.
for VCs, this is an assertion method, for VPs this would be authentication method.

hope it's alright to keep it in limbo for now.

internal/keyaccess/dataintegrity.go
decentralgabe: Yeah, this really returns a Provable.

I could split this into three like : SignVerifiableCredential, SignVerifiablePresentation, and just SignProvable as utilities but they will all do the same thing.

internal/keyaccess/jwt.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func (ka JWKKeyAccess) SignVerifiablePresentation(presentation credential.VerifiablePresentation) ([]byte, error) {
+	if err := presentation.IsValid(); err != nil {
+		return nil, errors.New("cannot sign invalid presentation")
+	}
+	return signing.SignVerifiablePresentationJWT(ka.JWTSigner, presentation)
+}
+
+func (ka JWKKeyAccess) VerifyVerifiablePresentation(token string) (*credential.VerifiablePresentation, error) {
```
nitro-neal, line 80: Same thing with token, eventually we could wrap this too

nitro-neal (approved):
🔥

internal/keyaccess/jwt.go
decentralgabe: yeah, this is a good point. there would not be anything enforcing it really but it is another level of thought that needs to go into the design. 

let me incorporate this into the next round.

PR #111, opened 2022-09-24
provide development builds as artifacts

pdxjohnny:
Via upload of built binaries post validation

Downstream consumers of ssi-service may wish to use bleeding edge functionality
without having easy access to golang build tooling. By providing binaries as an
option users do not have to worry about development environment testing to
start working with ssi-service latest features / bug fixes

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -26,8 +26,14 @@ var (
 
 // Build builds the library.
 func Build() error {
-	fmt.Println("Building...")
+	fmt.Println("Building library...")
 	return sh.Run(Go, "build", "-tags", "jwx_es256k", "./...")
+}
+
+// Artifact builds the binary.
+func Artifact() error {
+	fmt.Println("Building binary...")
+	return sh.Run(Go, "build", "-tags", "jwx_es256k", "-o", "./bin/ssi-service", "./cmd")
 }
 
 // Vuln downloads and runs govulncheck https://go.dev/blog/vuln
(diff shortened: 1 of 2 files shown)


pdxjohnny:
I'm not really sure where the binaries get built too, I was planning on using the first CI run to verify if that's the right path

pdxjohnny:
Okay pretty sure this should do it

decentralgabe (changes requested):
thanks @pdxjohnny can you add the upload artifacts to a new step separate from build, and put it under a mage command like `mage artifact` or similar?

also curious what the use is for having an artifact produced here?

michaelneale:
I guess the benefit is you have a versioned binary you can download vis the github CDN etc?

michaelneale:
but then they would have to run github actions locally etc. if its publishing binaries, ideally only want that done from trusted source (the github action server side).

magefile.go
```diff
@@ -24,12 +24,18 @@ var (
 	Go = "go"
 )
 
-// Build builds the library.
+// Build builds the package.
 func Build() error {
-	fmt.Println("Building...")
+	fmt.Println("Building package...")
```
pdxjohnny, line 29:
```suggestion
	fmt.Println("Building library...")
```

pdxjohnny:
Thanks for the comments! @michaelneale was spot on, as development here evolves I'd love to be able to pull from the latest stuff ASAP, without the need to replicate the build infrastructure (since I'm hoping to use this in many projects).

decentralgabe (approved):

decentralgabe:
yes, was referring to what you've included - building the artifact, not publishing it

PR #115, opened 2022-09-29
have dwn go through the svc endpoint

decentralgabe:
should have caught this earlier. really we should think of the 'service' apis as all that are exposed to other services. storage should be internal to each service

### DIFF
--- a/pkg/service/dwn/dwn.go
+++ b/pkg/service/dwn/dwn.go
@@ -3,22 +3,23 @@ package dwn
 import (
 	"fmt"
 
+	sdkutil "github.com/TBD54566975/ssi-sdk/util"
 	"github.com/sirupsen/logrus"
 
 	"github.com/tbd54566975/ssi-service/config"
 	"github.com/tbd54566975/ssi-service/internal/util"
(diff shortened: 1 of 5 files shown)


pkg/service/dwn/dwn.go
```diff
@@ -3,61 +3,69 @@ package dwn
...
+		manifest: manifest,
 	}, nil
 }
 
 func (s Service) GetManifest(request DWNPublishManifestRequest) (*DWNPublishManifestResponse, error) {
+
 	logrus.Debugf("getting manifest: %s", request.ManifestID)
 
-	gotManifest, err := s.manifestStorage.GetManifest(request.ManifestID)
+	gotManifest, err := s.manifest.GetManifest(manifest.GetManifestRequest{ID: request.ManifestID})
```
nitro-neal, line 68: Yup this is exactly what I was thinking about doing for a clean up 2nd pass.

We can even delete the dwn/model.go file and use CreateManifestRequest in manifest too, but that may be too much mixing

nitro-neal (approved):

PR #118, opened 2022-10-01
SIP3: Support Signing of Credentials

decentralgabe:
- Supports signing credentials with known keys, stored in key storage.
- Manifest and dependent APIs, tests, have been updated accordingly.
- Split out existing server tests to their own files for clarity

Separate verification APIs will come in a follow-up PR.


pkg/server/router/keystore.go
```diff
@@ -42,11 +42,15 @@ func (sk StoreKeyRequest) ToServiceRequest() (*keystore.StoreKeyRequest, error)
...
+	privKey, err := crypto.BytesToPrivKey(privateKeyBytes, sk.Type)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not convert bytes to private key")
+	}
 	return &keystore.StoreKeyRequest{
 		ID:         sk.ID,
 		Type:       sk.Type,
 		Controller: sk.Controller,
-		Key:        privateKeyBytes,
+		Key:        privKey,
```
nitro-neal, line 53: hopefully we wont need the privateKeyBytes for anything, this is a good forcing function for us to try not and pass around bytes

pkg/server/router/manifest.go
```diff
@@ -181,19 +180,20 @@ func (mr ManifestRouter) DeleteManifest(ctx context.Context, w http.ResponseWrit
...
 		Application:  sar.Application,
-		RequesterDID: sar.RequesterDID,
+		ApplicantDID: sar.ApplicantDID,
 	}
 }
 
 type SubmitApplicationResponse struct {
-	Response    manifestsdk.CredentialResponse    `json:"response"`
-	Credentials []credential.VerifiableCredential `json:"credentials"`
+	Response manifestsdk.CredentialResponse `json:"credential_response"`
```
nitro-neal, line 194: Other responses are camelCase right?

pkg/server/router/manifest.go
```diff
@@ -181,19 +180,20 @@ func (mr ManifestRouter) DeleteManifest(ctx context.Context, w http.ResponseWrit
 type SubmitApplicationRequest struct {
 	Application manifestsdk.CredentialApplication `json:"application" validate:"required"`
 	// Once we have JWT signed wrapper that can get the did this can be removed
```
nitro-neal, line 182: Is this comment still valid?

pkg/server/router/manifest.go
```diff
@@ -223,7 +223,16 @@ func (mr ManifestRouter) SubmitApplication(ctx context.Context, w http.ResponseW
 		return framework.NewRequestError(errors.Wrap(err, errMsg), http.StatusInternalServerError)
 	}
 
-	resp := SubmitApplicationResponse{Response: submitApplicationResponse.Response, Credentials: submitApplicationResponse.Credential}
+	var credentials []interface{}
+	for _, container := range submitApplicationResponse.Credential {
+		if container.HasDataIntegrityCredential() {
+			credentials = append(credentials, *container.Credential)
+		}
```
nitro-neal, line 230: should there be an else if? There could be a case where someone fills out a JWT and a DataIntegrityCred and it would print 2 credentials

pkg/service/credential/storage/storage.go
```diff
@@ -5,21 +5,45 @@ import (
...
+	CredentialJWT *string                          `json:"token,omitempty"`
+
+	Issuer       string `json:"issuer"`
+	Subject      string `json:"subject"`
+	Schema       string `json:"schema"`
+	IssuanceDate string `json:"issuanceDate"`
+}
+
+func (sc StoredCredential) IsValid() bool {
+	return sc.Credential != nil || sc.CredentialJWT != nil
```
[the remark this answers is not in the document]
nitro-neal: Or maybe better like our code should never be in this state, so an assertion or something

pkg/service/credential/model.go
```diff
@@ -45,7 +47,7 @@ type GetCredentialBySchemaRequest struct {
 }
 
 type GetCredentialsResponse struct {
-	Credentials []credsdk.VerifiableCredential
```
nitro-neal, line 48: If this credential.CredentialsContainer becomes confusing or problematic in the future we could go another route. Basically in the SDK like you can inittilize the oebject with a JWT or a normal json parse, and it would be a []credsdk.VerifiableCredential (but with some jwt handling extra fields if it was a jwt?)

nitro-neal (approved):

pkg/server/router/manifest.go
decentralgabe: this is tough because it is required by the spec: https://identity.foundation/credential-manifest/#embed-locations-2

pkg/service/credential/model.go
decentralgabe: let me take this suggestion

pkg/server/router/manifest.go
decentralgabe: it is, this will need to be changed to accept a verifiable presentation or jwt. will handle separately

pkg/server/router/manifest.go
decentralgabe: good call, updated

pkg/service/credential/storage/storage.go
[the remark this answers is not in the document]
decentralgabe: updated this

PR #121, opened 2022-10-07
Credential Verification & DID Resolution

decentralgabe:
adding support for credential verification, which has a dependency on DID resolution
APIs for both, and tests

### DIFF
--- a/internal/credential/verification.go
+++ b/internal/credential/verification.go
@@ -1,0 +1,1 @@
+package credential
--- a/internal/credential/verification.go
+++ b/internal/credential/verification.go
@@ -1,1 +1,161 @@
 package credential
+
+import (
+	"crypto"
+	"fmt"
+
+	credsdk "github.com/TBD54566975/ssi-sdk/credential"
+	"github.com/TBD54566975/ssi-sdk/credential/signing"
+	"github.com/TBD54566975/ssi-sdk/credential/verification"
+	didsdk "github.com/TBD54566975/ssi-sdk/did"
+	"github.com/goccy/go-json"
+	"github.com/lestrrat-go/jwx/jwk"
+	"github.com/mr-tron/base58"
+	"github.com/multiformats/go-multibase"
+	"github.com/multiformats/go-varint"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+
+	"github.com/tbd54566975/ssi-service/internal/keyaccess"
+	"github.com/tbd54566975/ssi-service/internal/util"
+)
+
+type CredentialVerifier struct {
+	verifier *verification.CredentialVerifier
+	resolver *didsdk.Resolver
+}
+
+// NewCredentialVerifier creates a new credential verifier which executes both signature and static verification checks.
+// In the future the set of verification checks will be configurable.
+func NewCredentialVerifier(resolver *didsdk.Resolver) (*CredentialVerifier, error) {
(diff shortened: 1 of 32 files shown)


internal/credential/verification.go
```diff
@@ -0,0 +1,176 @@
...
+)
+
+type CredentialVerifier struct {
+	verifier *verification.CredentialVerifier
+	resolver *didsdk.Resolver
+}
+
+// NewCredentialVerifier creates a new credential verifier which executes both signature and static verification checks.
+// In the future the set of verification checks will be configurable.
+func NewCredentialVerifier(resolver *didsdk.Resolver) (*CredentialVerifier, error) {
```
nitro-neal, line 31: eventually we will get a whole test suit setup with a lot of real deal signed creds and verify theses functions

internal/credential/verification.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	kid, pubKey, err := v.getVerificationInformation(cred.Issuer.(string), "")
+	if err != nil {
+		return util.LoggingErrorMsg(err, "could not get verification from credential")
+	}
+	verifier, err := keyaccess.NewJWKKeyAccessVerifier(kid, pubKey)
+	if err != nil {
+		return util.LoggingErrorMsg(err, "could not create verifier")
+	}
+	if err := verifier.Verify(keyaccess.JWKToken{Token: token}); err != nil {
```
nitro-neal, line 60: Now I can see wrapping with a JWKToken{} object instead of just using a string may become a nucance

pkg/service/credential/credential.go
```diff
@@ -168,6 +175,51 @@ func (s Service) signCredentialJWT(issuer string, cred credential.VerifiableCred
...
+// 4. If the credential has a schema, makes sure its data complies with the schema
+// LATER: Makes sure the credential has not been revoked, other checks.
+// Note: https://github.com/TBD54566975/ssi-sdk/issues/213
+func (s Service) VerifyCredential(request VerifyCredentialRequest) (*VerifyCredentialResponse, error) {
+	if err := request.IsValid(); err != nil {
+		return nil, util.LoggingErrorMsg(err, "invalid verify credential request")
+	}
+
+	if request.CredentialJWT != nil {
+		if err := s.verifier.VerifyJWTCredential(*request.CredentialJWT); err != nil {
```
nitro-neal, line 211: Could be cool for audit logging or dev sanity to have a function (or debug flag) that would log how many validatioons were run.

Would be terrible if only expiration date validator was applied to this and then we think its 'fully verified' but only a portion.

Maybe there should be a 'FullyValidate' function that runs all validators ?

nitro-neal (approved):
Goob job! Lots of complex stuff in here

internal/credential/verification.go
decentralgabe: agreed

pkg/service/credential/credential.go
decentralgabe: let me add the logging and think about fully validate

decentralgabe: the verify function is 'fully validate' maybe the name is confusing

as per your other comment I opened an issue here https://github.com/TBD54566975/ssi-service/issues/122

internal/credential/verification.go
decentralgabe: agree

PR #123, opened 2022-10-10
Schema Signing & Verification

decentralgabe:
- schema signing, verification functionality + APIs
- added delete schema apis 
- tests for all
- not yet integrated into credential creation code

### DIFF
--- a/pkg/server/router/did.go
+++ b/pkg/server/router/did.go
@@ -87,8 +87,15 @@ func (dr DIDRouter) CreateDIDByMethod(ctx context.Context, w http.ResponseWriter
 	}
 
 	var request CreateDIDByMethodRequest
+	invalidCreateDIDRequest := "invalid create DID request"
 	if err := framework.Decode(r, &request); err != nil {
-		errMsg := "invalid create DID request"
+		errMsg := invalidCreateDIDRequest
+		logrus.WithError(err).Error(errMsg)
+		return framework.NewRequestError(errors.Wrap(err, errMsg), http.StatusBadRequest)
+	}
+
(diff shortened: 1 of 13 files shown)


internal/keyaccess/verification.go
```diff
@@ -0,0 +1,108 @@
+package keyaccess
```
decentralgabe, line 1: moved this to its own package, since it has use outside of just cred verification

pkg/server/router/did.go
```diff
@@ -87,8 +87,15 @@ func (dr DIDRouter) CreateDIDByMethod(ctx context.Context, w http.ResponseWriter
 	}
 
 	var request CreateDIDByMethodRequest
+	invalidCreateDIDRequest := "invalid create DID request"
 	if err := framework.Decode(r, &request); err != nil {
```
nitro-neal, line 91: alternativly you could do 

if framework.Decode(r, &request) != nil || framework.ValidateRequest(request) != nil {
    errMsg := invalidCreateDIDRequest
		logrus.WithError(err).Error(errMsg)
		return framework.NewRequestError(errors.Wrap(err, errMsg), http.StatusBadRequest)
}

since its the same output (but you need to get err variable in a golang way somehow lol maybe not possible)

nitro-neal (approved):

pkg/server/router/did.go
decentralgabe: yeah, unfortunately this wouldn't work because both return unique errors and we couldn't capture both. conditional error assignment would be a nice feature 🤔

PR #126, opened 2022-10-11
Support schema usage in the credential service

decentralgabe:
- allow referencing schemas from the cred service
- streamlined service status / init code


decentralgabe:
the manifest test will fail until this is resolved and uptaken in the ssi-sdk, which needs to be uptaken here https://github.com/decentralized-identity/credential-manifest/pull/122

decentralgabe:
Relies on uptake in https://github.com/TBD54566975/ssi-sdk/pull/224

pkg/service/did/did.go
```diff
@@ -27,11 +28,40 @@ type Service struct {
...
-type MethodHandler interface {
-	CreateDID(request CreateDIDRequest) (*CreateDIDResponse, error)
-	GetDID(request GetDIDRequest) (*GetDIDResponse, error)
-	GetDIDs(method didsdk.Method) (*GetDIDsResponse, error)
+func (s *Service) Type() framework.Type {
+	return framework.DID
+}
+
+// Status is a self-reporting status for the DID service.
+func (s *Service) Status() framework.Status {
```
nitro-neal, line 36: for these status stuff you could potentially loop through the service object fields using reflection and it could probably be less code, could almost make just one helper function
https://stackoverflow.com/questions/18926303/iterate-through-the-fields-of-a-struct-in-go

decentralgabe: you are correct, and this may be a good change. the only thing that would be missing is the relative error messages, but perhaps this is ok and we can surface the struct name.

let me open an issue for this. https://github.com/TBD54566975/ssi-service/issues/127

nitro-neal (approved):
Awesome, I will update the 'getting started' to add a schema

PR #130, opened 2022-10-14
Credential Manifest signing and verification

decentralgabe:
- cred manifest signing and verification
- added json tags for required fields to all svc structs
- fixed the schema validator to do some static validation too
- updated the JWT type to be an alias not a struct, using this everywhere
- cred application signing coming next


pkg/service/manifest/manifest.go
```diff
@@ -65,34 +74,129 @@ func NewManifestService(config config.ManifestServiceConfig, s storage.ServiceSt
...
+	response := CreateManifestResponse{Manifest: m, ManifestJWT: *manifestJWT}
 	return &response, nil
 }
 
+func (s Service) signManifestJWT(m manifest.CredentialManifest) (*keyaccess.JWT, error) {
+	issuerID := m.Issuer.ID
+	gotKey, err := s.keyStore.GetKey(keystore.GetKeyRequest{ID: issuerID})
+	if err != nil {
+		errMsg := fmt.Sprintf("could not get key for signing manifest with key<%s>", issuerID)
+		return nil, util.LoggingErrorMsg(err, errMsg)
```
nitro-neal, line 120: errors.wrap vs util.LoggingErrorMsg(err, errMsg)
Eventually we will need to go through code and standardize, I'm guilty of this too we have 5 different ways to return errors

nitro-neal (approved):
looks good

pkg/service/manifest/manifest.go
decentralgabe: good catch. yeah, we need to have some rules on where we log vs where we return errors

definitely on the caller site but maybe we should be logging in each layer if there are more specific errors. like, in storage log specific storage errors, in the app layer log generation/modification errors, service layer something more generic...

PR #138, opened 2022-10-19
Update kyc manifest and credential test data

mistermoe:
@decentralgabe does `filter` support `required`?


decentralgabe:
> @decentralgabe does `filter` support `required`?

it supports any valid json schema - so this should be fine https://identity.foundation/presentation-exchange/#presentation-definition-plain-object
whether the library handles it correctly is another question...may be worth having a test against

decentralgabe (approved):

nitro-neal (approved):

PR #149, opened 2022-10-28
Enabled golangci-lint with staticcheck only and fixed issues

andresuribe87:
This PR configures and enables golangci-lint in github actions. I intentionally only enabled the `staticcheck` linter, so this becomes easier to review. The only noteworthy change to highlight was the due to the following warning: 

```
pkg/server/router/dwn.go:82:19: SA4022: the address of a variable cannot be nil (staticcheck)
        if err != nil || &publishManifestResponse.Manifest == nil {
```

To fix, I changed this to use the `IsEmpty()` method.

Additional linters will be enabled in a follow up PR.


pkg/server/router/dwn.go
```diff
@@ -46,8 +46,8 @@ func (req PublishManifestRequest) ToServiceRequest() dwn.PublishManifestRequest
 }
 
 type PublishManifestResponse struct {
-	Manifest    manifest.CredentialManifest       `json:"manifest" validate:"required"`
-	DWNResponse dwnpkg.DWNPublishManifestResponse `json:"dwnResponse" validate:"required"`
+	Manifest    *manifest.CredentialManifest       `json:"manifest" validate:"required"`
+	DWNResponse *dwnpkg.DWNPublishManifestResponse `json:"dwnResponse" validate:"required"`
```
decentralgabe, line 50: i'd prefer reverting this, required fields should not be nil-able

pkg/server/router/dwn.go
```diff
@@ -79,20 +79,20 @@ func (dwnr DWNRouter) PublishManifest(ctx context.Context, w http.ResponseWriter
 	req := request.ToServiceRequest()
 	publishManifestResponse, err := dwnr.service.GetManifest(req)
 
-	if err != nil || &publishManifestResponse.Manifest == nil {
+	if err != nil || publishManifestResponse.Manifest == nil {
```
decentralgabe, line 82: to get around this check we should define an `IsEmpty` method on Manifest. I think one may even already exist.

decentralgabe (changes requested):

pkg/server/router/dwn.go
andresuribe87: Done.

pkg/server/router/dwn.go
andresuribe87: Sounds good.

decentralgabe (approved):

PR #159, opened 2022-11-03
Enabled a new set of linters

andresuribe87:
Most should be self-explanatory.

### DIFF
--- a/internal/did/resolver.go
+++ b/internal/did/resolver.go
@@ -12,7 +12,7 @@ func BuildResolver(methods []string) (*didsdk.Resolver, error) {
 	if len(methods) == 0 {
 		return nil, errors.New("no methods provided")
 	}
-	var resolvers []didsdk.Resolution
+	resolvers := make([]didsdk.Resolution, 0, len(methods))
 	for _, method := range methods {
 		resolver, err := getKnownResolver(method)
 		if err != nil {
--- a/pkg/server/framework/request.go
+++ b/pkg/server/framework/request.go
@@ -69,8 +69,7 @@ func Decode(r *http.Request, val interface{}) error {
 
 	if err := validate.Struct(val); err != nil {
 		var vErrors validator.ValidationErrors
-		ok := errors.As(err, &vErrors)
-		if !ok {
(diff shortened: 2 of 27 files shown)


internal/did/resolver.go
```diff
@@ -12,7 +12,7 @@ func BuildResolver(methods []string) (*didsdk.Resolver, error) {
 	if len(methods) == 0 {
 		return nil, errors.New("no methods provided")
 	}
-	var resolvers []didsdk.Resolution
+	resolvers := make([]didsdk.Resolution, 0, len(methods))
```
decentralgabe, line 15: will this end up with extra zero-ed out values if not all methods get added?

pkg/server/framework/request.go
```diff
@@ -68,7 +68,8 @@ func Decode(r *http.Request, val interface{}) error {
 	}
 
 	if err := validate.Struct(val); err != nil {
-		vErrors, ok := err.(validator.ValidationErrors)
+		var vErrors validator.ValidationErrors
+		ok := errors.As(err, &vErrors)
 		if !ok {
 			return err
 		}
```
decentralgabe, line 75:
```suggestion
		if ok := errors.As(err, &vErrors); !ok {
			return err
		}
```

pkg/server/framework/response.go
```diff
@@ -43,7 +43,9 @@ func Respond(ctx context.Context, w http.ResponseWriter, data interface{}, statu
 func RespondError(ctx context.Context, w http.ResponseWriter, err error) error {
 	// if the cause of the error provided is a `SafeError`, construct an ErrorResponse
 	// using the contents of SafeError and send it back to the client
-	if webErr, ok := errors.Cause(err).(*SafeError); ok {
+	var webErr *SafeError
+	ok := errors.As(errors.Cause(err), &webErr)
+	if ok {
```
decentralgabe, line 48:
```suggestion
	if ok := errors.As(errors.Cause(err), &webErr); ok {
```

decentralgabe (approved):

PR #161, opened 2022-11-04
[OSE-178] Added /v1/presentation/definition endpoint

andresuribe87:
This PR adds the the endpoint with Create, Get, and Delete operations. 

Unit tests were also added to ensure things work correctly.

This is part of supporting the [Presentation Exchange](https://identity.foundation/presentation-exchange/#presentation-definition) spec.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -48,11 +48,12 @@ type ServicesConfig struct {
 
 	// Embed all service-specific configs here. The order matters: from which should be instantiated first, to last
 
-	KeyStoreConfig   KeyStoreServiceConfig   `toml:"keystore,omitempty"`
-	DIDConfig        DIDServiceConfig        `toml:"did,omitempty"`
-	SchemaConfig     SchemaServiceConfig     `toml:"schema,omitempty"`
-	CredentialConfig CredentialServiceConfig `toml:"credential,omitempty"`
-	ManifestConfig   ManifestServiceConfig   `toml:"manifest,omitempty"`
+	KeyStoreConfig     KeyStoreServiceConfig   `toml:"keystore,omitempty"`
+	DIDConfig          DIDServiceConfig        `toml:"did,omitempty"`
+	SchemaConfig       SchemaServiceConfig     `toml:"schema,omitempty"`
+	CredentialConfig   CredentialServiceConfig `toml:"credential,omitempty"`
+	ManifestConfig     ManifestServiceConfig   `toml:"manifest,omitempty"`
+	PresentationConfig PresentationServiceConfig `toml:"presentation,omitempty"`
 }
 
 // BaseServiceConfig represents configurable properties for a specific component of the SSI Service
@@ -122,6 +123,17 @@ func (m *ManifestServiceConfig) IsEmpty() bool {
 	return reflect.DeepEqual(m, &ManifestServiceConfig{})
 }
 
+type PresentationServiceConfig struct {
+	*BaseServiceConfig
+}
+
+func (d *PresentationServiceConfig) IsEmpty() bool {
+	if d == nil {
+		return true
+	}
+	return reflect.DeepEqual(d, &PresentationServiceConfig{})
+}
+
 // LoadConfig attempts to load a TOML config file from the given path, and coerce it into our object model.
 // Before loading, defaults are applied on certain properties, which are overwritten if specified in the TOML file.
 func LoadConfig(path string) (*SSIServiceConfig, error) {
@@ -182,6 +194,9 @@ func LoadConfig(path string) (*SSIServiceConfig, error) {
 			},
 			ManifestConfig: ManifestServiceConfig{
 				BaseServiceConfig: &BaseServiceConfig{Name: "manifest"},
+			},
+			PresentationConfig: PresentationServiceConfig{
+				BaseServiceConfig: &BaseServiceConfig{Name: "presentation"},
 			},
 		}
 	} else {
--- a/config/config.go
+++ b/config/config.go
@@ -48,11 +48,11 @@ type ServicesConfig struct {
 
 	// Embed all service-specific configs here. The order matters: from which should be instantiated first, to last
 
-	KeyStoreConfig     KeyStoreServiceConfig   `toml:"keystore,omitempty"`
-	DIDConfig          DIDServiceConfig        `toml:"did,omitempty"`
-	SchemaConfig       SchemaServiceConfig     `toml:"schema,omitempty"`
-	CredentialConfig   CredentialServiceConfig `toml:"credential,omitempty"`
-	ManifestConfig     ManifestServiceConfig   `toml:"manifest,omitempty"`
+	KeyStoreConfig     KeyStoreServiceConfig     `toml:"keystore,omitempty"`
+	DIDConfig          DIDServiceConfig          `toml:"did,omitempty"`
+	SchemaConfig       SchemaServiceConfig       `toml:"schema,omitempty"`
+	CredentialConfig   CredentialServiceConfig   `toml:"credential,omitempty"`
+	ManifestConfig     ManifestServiceConfig     `toml:"manifest,omitempty"`
 	PresentationConfig PresentationServiceConfig `toml:"presentation,omitempty"`
 }
 
@@ -127,11 +127,11 @@ type PresentationServiceConfig struct {
 	*BaseServiceConfig
 }
 
-func (d *PresentationServiceConfig) IsEmpty() bool {
-	if d == nil {
+func (p *PresentationServiceConfig) IsEmpty() bool {
+	if p == nil {
 		return true
 	}
-	return reflect.DeepEqual(d, &PresentationServiceConfig{})
+	return reflect.DeepEqual(p, &PresentationServiceConfig{})
 }
 
 // LoadConfig attempts to load a TOML config file from the given path, and coerce it into our object model.
--- a/pkg/server/router/presentation.go
+++ b/pkg/server/router/presentation.go
@@ -1,0 +1,140 @@
+package router
+
+import (
+	"context"
+	"fmt"
+	"github.com/TBD54566975/ssi-sdk/credential/exchange"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"github.com/tbd54566975/ssi-service/pkg/server/framework"
+	svcframework "github.com/tbd54566975/ssi-service/pkg/service/framework"
+	"github.com/tbd54566975/ssi-service/pkg/service/presentation"
+	"net/http"
+)
+
+type PresentationDefinitionRouter struct {
+	service *presentation.Service
+}
+
+func NewPresentationDefinitionRouter(s svcframework.Service) (*PresentationDefinitionRouter, error) {
+	if s == nil {
+		return nil, errors.New("service cannot be nil")
+	}
+	schemaService, ok := s.(*presentation.Service)
+	if !ok {
+		return nil, fmt.Errorf("could not create presentation router with service type: %s", s.Type())
+	}
+	return &PresentationDefinitionRouter{service: schemaService}, nil
+}
+
+type CreatePresentationDefinitionRequest struct {
+	PresentationDefinition exchange.PresentationDefinition `json:"presentation_definition" validate:"required"`
+}
+
+type CreatePresentationDefinitionResponse struct {
+	PresentationDefinition exchange.PresentationDefinition `json:"presentation_definition"`
+}
+
+// CreatePresentationDefinition godoc
+// @Summary      Create PresentationDefinition
+// @Description  Create presentation definition
+// @Tags         PresentationDefinitionAPI
+// @Accept       json
+// @Produce      json
+// @Param        request  body      CreatePresentationDefinitionRequest  true  "request body"
+// @Success      201      {object}  CreatePresentationDefinitionResponse
+// @Failure      400      {string}  string  "Bad request"
+// @Failure      500      {string}  string  "Internal server error"
+// @Router       /v1/schemas [put]
+func (sr PresentationDefinitionRouter) CreatePresentationDefinition(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
+	var request CreatePresentationDefinitionRequest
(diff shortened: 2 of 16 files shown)


pkg/server/router/presentation.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	service, ok := s.(*presentation.Service)
+	if !ok {
+		return nil, fmt.Errorf("could not create presentation router with service type: %s", s.Type())
+	}
+	return &PresentationDefinitionRouter{service: service}, nil
+}
+
+type CreatePresentationDefinitionRequest struct {
+	PresentationDefinition exchange.PresentationDefinition `json:"presentation_definition" validate:"required"`
```
nitro-neal, line 31: we have been doing camelCase
Change the json strings to presentationDefinition

pkg/server/router/presentation.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+type CreatePresentationDefinitionResponse struct {
+	PresentationDefinition exchange.PresentationDefinition `json:"presentation_definition"`
```
nitro-neal, line 35: same here

pkg/server/router/presentation.go
```diff
@@ -0,0 +1,140 @@
...
+// @Produce      json
+// @Param        request  body      CreatePresentationDefinitionRequest  true  "request body"
+// @Success      201      {object}  CreatePresentationDefinitionResponse
+// @Failure      400      {string}  string  "Bad request"
+// @Failure      500      {string}  string  "Internal server error"
+// @Router       /v1/presentation/definition [put]
+func (sr PresentationDefinitionRouter) CreatePresentationDefinition(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
+	var request CreatePresentationDefinitionRequest
+	if err := framework.Decode(r, &request); err != nil {
+		errMsg := "decoding"
```
nitro-neal, line 52: Fix the error message:
Invalid Presentation Definition Request 
or something like that

pkg/server/router/presentation.go
```diff
@@ -0,0 +1,140 @@
...
+		logrus.WithError(err).Error(errMsg)
+		return framework.NewRequestError(errors.Wrap(err, errMsg), http.StatusBadRequest)
+	}
+
+	if err := framework.ValidateRequest(request); err != nil {
+		errMsg := "validating"
```
nitro-neal, line 58: fix error message

pkg/server/router/presentation.go
```diff
@@ -0,0 +1,140 @@
...
+		return framework.NewRequestError(errors.Wrap(err, errMsg), http.StatusInternalServerError)
+	}
+
+	resp := CreatePresentationDefinitionResponse{def.PresentationDefinition}
+	return framework.Respond(ctx, w, resp, http.StatusCreated)
+}
+
+type GetPresentationDefinitionResponse struct {
+	ID                     string                          `json:"id"`
+	PresentationDefinition exchange.PresentationDefinition `json:"presentation_definition"`
```
nitro-neal, line 79: camelCase

pkg/server/server.go
```diff
@@ -26,6 +26,7 @@ const (
 	DIDsPrefix         = "/dids"
 	SchemasPrefix      = "/schemas"
 	CredentialsPrefix  = "/credentials"
+	PresentationPrefix = "/presentation"
```
nitro-neal, line 29: should be presentations

pkg/service/presentation/storage/storage.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"fmt"
+	"github.com/TBD54566975/ssi-sdk/credential/exchange"
+	"github.com/tbd54566975/ssi-service/internal/util"
+	"github.com/tbd54566975/ssi-service/pkg/storage"
+)
+
+type StoredPresentation struct {
+	ID                     string                          `json:"id"`
+	PresentationDefinition exchange.PresentationDefinition `json:"presentation_definition"`
```
nitro-neal, line 12: camelCase

pkg/service/presentation/model.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/TBD54566975/ssi-sdk/credential/exchange"
+	"github.com/TBD54566975/ssi-sdk/util"
+)
+
+const (
+	Version1 string = "1.0"
+)
+
+type CreatePresentationDefinitionRequest struct {
+	PresentationDefinition exchange.PresentationDefinition `json:"presentation_definition" validate:"required"`
```
nitro-neal, line 13: camel

pkg/service/presentation/service.go
```diff
@@ -0,0 +1,109 @@
...
+		return nil, util.LoggingErrorMsg(err, "could not store presentation")
+	}
+
+	return &CreatePresentationDefinitionResponse{
+		PresentationDefinition: storedPresentation.PresentationDefinition,
+	}, nil
+}
+
+func (s Service) GetPresentationDefinition(request GetPresentationDefinitionRequest) (*GetPresentationDefinitionResponse, error) {
+	logrus.Debugf("getting presentation definition: %s", request.ID)
```
nitro-neal, line 86: are these debug logs on purpose?

pkg/service/presentation/model.go
```diff
@@ -0,0 +1,35 @@
+package presentation
+
+import (
+	"github.com/TBD54566975/ssi-sdk/credential/exchange"
+	"github.com/TBD54566975/ssi-sdk/util"
+)
+
+const (
+	Version1 string = "1.0"
```
nitro-neal, line 9: Where is this used?

config/config.go
```diff
@@ -122,6 +123,17 @@ func (m *ManifestServiceConfig) IsEmpty() bool {
 	return reflect.DeepEqual(m, &ManifestServiceConfig{})
 }
 
+type PresentationServiceConfig struct {
+	*BaseServiceConfig
+}
+
+func (d *PresentationServiceConfig) IsEmpty() bool {
```
decentralgabe, line 130: nit: `p *PresentationServiceConfig`

pkg/server/router/presentation.go
decentralgabe: I do not think we should take in an entire presentation definition here, instead we should take in what is necessary to build one. this is similar to our manifest and credential apis today:

see: https://github.com/TBD54566975/ssi-service/blob/main/pkg/server/router/manifest.go#L43

pkg/server/router/presentation.go
decentralgabe: this response should be similar to https://github.com/TBD54566975/ssi-service/blob/main/pkg/server/router/manifest.go#L65

which does return the snake_case version because that is how the spec defines the property.

should this also return the JWT, or will you be handling that in another PR?

pkg/server/router/presentation.go
```diff
@@ -0,0 +1,140 @@
...
+// @Description  Create presentation definition
+// @Tags         PresentationDefinitionAPI
+// @Accept       json
+// @Produce      json
+// @Param        request  body      CreatePresentationDefinitionRequest  true  "request body"
+// @Success      201      {object}  CreatePresentationDefinitionResponse
+// @Failure      400      {string}  string  "Bad request"
+// @Failure      500      {string}  string  "Internal server error"
+// @Router       /v1/presentation/definition [put]
+func (sr PresentationDefinitionRouter) CreatePresentationDefinition(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
```
decentralgabe, line 49: nit: change `sr` to `pdr` or `pr`

pkg/server/server_presentation_test.go
```diff
@@ -0,0 +1,126 @@
...
+			assert.NoError(t, pRouter.GetPresentationDefinition(newRequestContextWithParams(map[string]string{"id": pd.ID}), w, req))
+
+			var resp router.GetPresentationDefinitionResponse
+			assert.NoError(t, json.NewDecoder(w.Body).Decode(&resp))
+			assert.Equal(t, pd.ID, resp.ID)
+			assert.Equal(t, *pd, resp.PresentationDefinition)
+
+			w.Flush()
+		}
+		{
```
decentralgabe, line 81: haven't seen this syntax before 🤔

pkg/service/presentation/model.go
decentralgabe: +1 we are using V2 of the spec or is this for something else?

pkg/service/presentation/model.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+type CreatePresentationDefinitionRequest struct {
+	PresentationDefinition exchange.PresentationDefinition `json:"presentation_definition" validate:"required"`
+}
+
+func (csr CreatePresentationDefinitionRequest) IsValid() bool {
```
decentralgabe, line 16: `cpr`?

pkg/service/presentation/service.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// CreatePresentationDefinition houses the main service logic for presentation definition creation. It validates the input, and
+// produces a presentation definition value that conforms with the PresentationDefinition specification.
+func (s Service) CreatePresentationDefinition(request CreatePresentationDefinitionRequest) (*CreatePresentationDefinitionResponse, error) {
+	logrus.Debugf("creating presentation definition: %+v", request)
+
+	if !request.IsValid() {
+		errMsg := fmt.Sprintf("invalid create presentation definition request: %+v", request)
+		return nil, util.LoggingNewError(errMsg)
```
decentralgabe, line 67: could use the `util. LoggingNewErrorf` method

pkg/service/presentation/service.go
```diff
@@ -0,0 +1,109 @@
...
+		PresentationDefinition: storedPresentation.PresentationDefinition,
+	}, nil
+}
+
+func (s Service) GetPresentationDefinition(request GetPresentationDefinitionRequest) (*GetPresentationDefinitionResponse, error) {
+	logrus.Debugf("getting presentation definition: %s", request.ID)
+
+	storedPresentation, err := s.storage.GetPresentation(request.ID)
+	if err != nil {
+		err := errors.Wrapf(err, "error getting presentation definition: %s", request.ID)
```
decentralgabe, line 90: this err is shadowed so you can use `=` instead of `:=`

pkg/service/presentation/service.go
```diff
@@ -0,0 +1,109 @@
...
+		return nil, util.LoggingError(err)
```
decentralgabe, line 91: consider `util.LoggingNewErrorf`

pkg/service/presentation/service.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	if storedPresentation == nil {
+		err := fmt.Errorf("presentation definition with id<%s> could not be found", request.ID)
+		return nil, util.LoggingError(err)
```
decentralgabe, line 95: same here

pkg/service/presentation/service.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	return &GetPresentationDefinitionResponse{ID: storedPresentation.ID, PresentationDefinition: storedPresentation.PresentationDefinition}, nil
+}
+
+func (s Service) DeletePresentationDefinition(request DeletePresentationDefinitionRequest) error {
+	logrus.Debugf("deleting presentation definition: %s", request.ID)
+
+	if err := s.storage.DeletePresentation(request.ID); err != nil {
+		errMsg := fmt.Sprintf("could not delete presentation definition with id: %s", request.ID)
+		return util.LoggingErrorMsg(err, errMsg)
```
decentralgabe, line 105: and here

pkg/service/presentation/storage/storage.go
decentralgabe: agree here, we only need to use snake_case as it is required by the specs in our API boundaries. internal to the system we should use camel. similar to - https://github.com/TBD54566975/ssi-service/blob/main/pkg/service/manifest/storage/storage.go#L12

pkg/service/presentation/storage/storage.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// NewPresentationStorage finds the presentation storage impl for a given ServiceStorage value
+func NewPresentationStorage(s storage.ServiceStorage) (Storage, error) {
+	switch s.Type() {
+	case storage.Bolt:
+		gotBolt, ok := s.(*storage.BoltDB)
+		if !ok {
+			errMsg := fmt.Sprintf("trouble instantiating : %s", s.Type())
+			return nil, util.LoggingNewError(errMsg)
```
decentralgabe, line 28: could be `util.LoggingNewErrorf`

thereisnogabe (changes requested):

pkg/server/router/presentation.go
andresuribe87: Done

andresuribe87 (commented):
Thanks for all the comments! PTAL

pkg/server/router/presentation.go
decentralgabe: wasn't sure if there was going to be a distinct endpoint for presentation requests. if there are - then disregard.

pkg/server/server_presentation_test.go
decentralgabe: nice!

pkg/service/presentation/service.go
decentralgabe: think they're fine we may want to add more

pkg/server/router/presentation.go
decentralgabe: For the UI this may push some logic into their code to build a compliant presentation definition instead of passing in the form fields and us building the object ourselves. Another factor at play is if we use the `Builder` methods we have in the sdk for each of these objects we can have higher certainty that we're using the latest supported object models and always building spec-compliant objects.

I can definitely see the appeal of the approach you presented - just have to consider the tradeoffs.

One other note - with an optional `keyID` we would need to create our own ID to map to theirs or else risk ID collisions.

thereisnogabe (approved):

PR #164, opened 2022-11-07
adding sip 7

nitro-neal:


decentralgabe (approved):
approving - looks like there's some extra spacing you will want to adjust before merging

PR #169, opened 2022-11-10
Adding credential revocation

nitro-neal:
This change allows you to create a credential that is revocable 

If you choose to ad  "revoked": true to the creation of a verifiable credential it will create a credential status

 ```
 "credentialStatus": {
            "id": "http://localhost:8080/v1/credentials/e70fd408-a00d-4908-b218-23695f74e18b/status",
            "statusListCredential": "http://localhost:8080/v1/credentials/status/6865a53c-c079-425b-a8bc-26ccc834021e",
            "statusListIndex": "91561",
            "statusPurpose": "revocation",
            "type": "StatusList2021Entry"
        },
```

This will create a statusListCredential for that existing issuer,schema pair

### CODE REVIEW

File: pkg/server/router/credential.go [hunk 1]
```diff
@@ -46,9 +46,10 @@ type CreateCredentialRequest struct {
 	// A context is optional. If not present, we'll apply default, required context values.
 	Context string `json:"@context"`
 	// A schema is optional. If present, we'll attempt to look it up and validate the data against it.
-	Schema string                 `json:"schema"`
-	Data   map[string]interface{} `json:"data" validate:"required"`
-	Expiry string                 `json:"expiry"`
+	Schema    string                 `json:"schema"`
+	Data      map[string]interface{} `json:"data" validate:"required"`
+	Expiry    string                 `json:"expiry"`
+	Revocable bool                   `json:"revocable"`
```
andresuribe87, line 52: This might be part of a bigger conversation, but do we add documentation to the fields that we're adding? API consumers will need to understand what every field does when they make a request, and not all are self explanatory.

pkg/server/router/credential.go
```diff
@@ -145,6 +147,140 @@ func (cr CredentialRouter) GetCredential(ctx context.Context, w http.ResponseWri
 	return framework.Respond(ctx, w, resp, http.StatusOK)
 }
 
+type GetCredentialStatusResponse struct {
+	Revoked bool `json:"revoked"`
```
andresuribe87, line 151: I remember having a conversation around `suspendable` credentials. What was the final decision on that?

pkg/server/router/credential.go
```diff
@@ -145,6 +147,140 @@ func (cr CredentialRouter) GetCredential(ctx context.Context, w http.ResponseWri
...
+
+	resp := GetCredentialStatusResponse{
+		Revoked: gotCredential.Revoked,
+	}
+
+	return framework.Respond(ctx, w, resp, http.StatusOK)
+}
+
+type GetCredentialStatusListResponse struct {
+	ID            string                        `json:"id"`
```
andresuribe87, line 187: The credential object already has a field called `ID`. What's the reasoning behind including both?

pkg/server/router/credential.go
```diff
@@ -145,6 +147,140 @@ func (cr CredentialRouter) GetCredential(ctx context.Context, w http.ResponseWri
...
+
+	return framework.Respond(ctx, w, resp, http.StatusOK)
+}
+
+type UpdateCredentialStatusRequest struct {
+	ID      string `json:"id" validate:"required"`
+	Revoked bool   `json:"revoked" validate:"required"`
+}
+
+func (c UpdateCredentialStatusRequest) ToServiceRequest() credential.UpdateCredentialStatusRequest {
```
andresuribe87, line 231: I don't think this needs to be exported? 
```suggestion
func (c UpdateCredentialStatusRequest) toServiceRequest() credential.UpdateCredentialStatusRequest {
```

pkg/server/router/credential.go
```diff
@@ -145,6 +147,140 @@ func (cr CredentialRouter) GetCredential(ctx context.Context, w http.ResponseWri
...
+		ID:            gotCredential.ID,
+		Credential:    gotCredential.Credential,
+		CredentialJWT: gotCredential.CredentialJWT,
+	}
+
+	return framework.Respond(ctx, w, resp, http.StatusOK)
+}
+
+type UpdateCredentialStatusRequest struct {
+	ID      string `json:"id" validate:"required"`
```
andresuribe87, line 227: Would it make sense to remove this field, given that the path will already have it?

pkg/server/router/credential_test.go
```diff
@@ -194,4 +201,97 @@ func TestCredentialRouter(t *testing.T) {
...
+					"type": "string",
+				},
+			},
+			"required":             []interface{}{"email"},
+			"additionalProperties": false,
+		}
+
+		createdSchema, err := schemaService.CreateSchema(schema.CreateSchemaRequest{Author: "me", Name: "simple schema", Schema: emailSchema})
+		assert.NoError(tt, err)
+		assert.NotEmpty(tt, createdSchema)
```
andresuribe87, line 247: Is this needed/interesting for the behavior that you're testing?

pkg/server/router/credential_test.go
```diff
@@ -194,4 +201,97 @@ func TestCredentialRouter(t *testing.T) {
 		assert.Error(tt, err)
 		assert.Contains(tt, err.Error(), fmt.Sprintf("credential not found with id: %s", cred.ID))
 	})
+
+	t.Run("Credential Status List Test", func(tt *testing.T) {
```
andresuribe87, line 205: I typically advocate for writing unit tests that test a single behavior. This is also known as [avoiding multiple acts](https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices#avoid-multiple-acts). In this case, I'd recommend separating this into multiple different behaviors that you're testing.

I know this isn't the style that we've been following for tests, so I'm ok to leave for consistency; but I do think it makes for tests that are easier to read, and thus maintain.

pkg/server/server_credential_test.go
```diff
@@ -621,4 +621,330 @@ func TestCredentialAPI(t *testing.T) {
 		assert.False(tt, verifyResp.Verified)
 		assert.Contains(tt, verifyResp.Reason, "could not parse credential from JWT")
 	})
+
+	t.Run("Test Create Revocable Credential", func(tt *testing.T) {
```
andresuribe87, line 625: Here and elsewhere. It's easier to read tests when the names include the:
* The name of the method being tested.
* The scenario under which it's being tested.
* The expected behavior when the scenario is invoked.

I found it difficult to do that for this test. That's typically a symptom that it's testing multiple behaviors, so maybe it's an opportunity to split it into smaller units. Alternatively, we can leave these tests here to act more like integration tests, which are following more a user story, rather than testing a behavior. I'd personally vouch for more, smaller, and targeted unit tests. That makes maintaining them much easier.

pkg/service/credential/service.go
```diff
@@ -338,6 +412,139 @@ func (s Service) GetCredentialsBySchema(request GetCredentialBySchemaRequest) (*
...
+	response := GetCredentialStatusResponse{
+		Revoked: gotCred.Revoked,
+	}
+	return &response, nil
+}
+
+func (s Service) GetCredentialStatusList(request GetCredentialStatusListRequest) (*GetCredentialStatusListResponse, error) {
+	logrus.Debugf("getting credential status list: %s", request.ID)
+
+	// statusListCredID := fmt.Sprintf("%s/v1/credentials/status/%s", s.config.ServiceEndpoint, request.ID)
```
andresuribe87, line 434: Why is this commented out?

pkg/service/credential/service.go
```diff
@@ -338,6 +412,139 @@ func (s Service) GetCredentialsBySchema(request GetCredentialBySchemaRequest) (*
...
+
+	gotCred, err := s.storage.GetCredential(request.ID)
+	if err != nil {
+		return nil, util.LoggingErrorMsgf(err, "could not get credential: %s", request.ID)
+	}
+	if !gotCred.IsValid() {
+		return nil, util.LoggingNewErrorf("credential returned is not valid: %s", request.ID)
+	}
+
+	// if the request is the same as what the current credential is there is no action
```
andresuribe87, line 464: Personal style preference: I generally recommend not adding inline comments unless there is clarification needed for why the code is doing something. I.e. only comment on "why", not "how" (because the "how" is done by the code)

pkg/service/credential/storage/bolt.go
```diff
@@ -1,35 +1,138 @@
...
+	Index int `json:"index"`
+}
+
 func NewBoltCredentialStorage(db *storage.BoltDB) (*BoltCredentialStorage, error) {
 	if db == nil {
 		return nil, errors.New("bolt db reference is nil")
 	}
+
+	// TODO: (Neal) there is a current bug with our Bolt implementation where if we do a GET without anything in the db it will throw an error
+	// Doing initial writes and then deleting will "warm up" our database and when we do a GET after that it will not crash and return empty list
```
andresuribe87, line 39: This is incredibly suspect and I strongly suspect there is another error going around.

pkg/service/credential/storage/bolt.go
```diff
@@ -1,35 +1,138 @@
...
+	}
+
 	return &BoltCredentialStorage{db: db}, nil
 }
 
+func (b BoltCredentialStorage) GetNextStatusListRandomIndex() (int, error) {
+
+	gotUniqueNumBytes, err := b.db.Read(statusListIndexNamespace, "status-list-indexes")
+	if err != nil {
+		return -1, util.LoggingErrorMsgf(err, "could not get unique numbers from db")
```
andresuribe87, line 92: I typically recommend having more succinct error messages, that don't include synonyms of error (e.g. "could not", "failed", "break", "fault").

The result is logs that are easier to read, and are not repetitive. Compare the following two log lines:
* `ERROR: could not get unique numbers from db: db not initialized`
* `ERROR: reading status list: db not initialized`.

The second conveys the same meaning, but is easier to read and avoids repetition.

```suggestion
		return -1, util.LoggingErrorMsgf(err, "reading status list")
```

config/compose.toml
```diff
@@ -37,6 +37,7 @@ name = "schema"
 
 [services.credential]
 name = "credential"
+service_endpoint = "http://localhost:8080"
```
decentralgabe, line 40: should this be in another section? ideally this is provided once for the entire server not for each service

internal/credential/model.go
```diff
@@ -19,6 +19,7 @@ type Container struct {
 	ID            string
 	Credential    *credential.VerifiableCredential
 	CredentialJWT *keyaccess.JWT
+	Revoked       bool
```
decentralgabe, line 22: I'm wondering if we should only expose `revoked` or allow this to be a `map[string]bool` of all possible statuses a credential can have. possible a later improvement.

pkg/server/router/credential.go
decentralgabe: I'm not sure how this ties into the swagger generation we have today. It's a good note - would you mind opening an issue to improve the api docs?

pkg/server/router/credential.go
decentralgabe: aligns with my comment above. we can choose to support specific statuses like 'revoked' or 'suspended' or leave it open to the administrators to define the statuses they want to have present on each credential - which could be none, one, or many

pkg/server/router/credential.go
decentralgabe: if the credential comes back as a JWT this lets the caller know the ID without needing to decode the JWT

pkg/server/router/credential.go
decentralgabe: +1

pkg/server/router/credential_test.go
decentralgabe: agree with this our tests need some improvement. it could be good to start the practice now (and going forward) and open up some issues for refactoring our existing tests.

pkg/server/router/credential_test.go
decentralgabe: we could refactor this to have a setup method of some sort that takes in `t *testing.T` but uses `require....` statements instead of `assert` as these objects are preconditions for the interesting parts of the test

pkg/service/credential/service.go
```diff
@@ -146,6 +149,74 @@ func (s Service) CreateCredential(request CreateCredentialRequest) (*CreateCrede
 		return nil, util.LoggingErrorMsg(err, errMsg)
 	}
 
+	if request.Revocable == true {
```
decentralgabe, line 152:
```suggestion
	if request.Revocable {
```

pkg/service/credential/service.go
```diff
@@ -146,6 +149,74 @@ func (s Service) CreateCredential(request CreateCredentialRequest) (*CreateCrede
...
+		// This should never happen, there should always be only 1 status list credential per <issuer,schema> pair
+		if len(storedStatusListCreds) > 1 {
+			return nil, util.LoggingNewErrorf("only one status list credential per <issuer,schema> pair allowed. issuer: %s schema: %s", request.Issuer, request.JSONSchema)
+		}
+
+		var statusListCredentialID string
+
+		// First time that this <issuer,schema> pair has a revocation or suspension credential issued
+		if len(storedStatusListCreds) == 0 {
+			generatedStatusListCredential, err := statussdk.GenerateStatusList2021Credential(fmt.Sprintf("%s/v1/credentials/status/%s", s.config.ServiceEndpoint, uuid.New().String()), issuerID, statussdk.StatusRevocation, []credential.VerifiableCredential{})
```
decentralgabe, line 171: can you pull out the URI string to a new line for readability?

decentralgabe, line 171: may want to break this into a new helper method `createNewStatusListCredential` or similar

pkg/service/credential/service.go
```diff
@@ -146,6 +149,74 @@ func (s Service) CreateCredential(request CreateCredentialRequest) (*CreateCrede
...
+
+			if err = s.storage.StoreStatusListCredential(storageRequest); err != nil {
+				return nil, util.LoggingErrorMsg(err, "could not store credential")
+			}
+
+		} else {
+			statusListCredentialID = storedStatusListCreds[0].ID
+		}
+
+		statusListIndex, err := s.storage.GetNextStatusListRandomIndex()
```
decentralgabe, line 201: may want to put this in an `UpdateCredentialStatusListCredential` or similar

pkg/service/credential/service.go
```diff
@@ -146,6 +149,74 @@ func (s Service) CreateCredential(request CreateCredentialRequest) (*CreateCrede
...
+			ID:                   fmt.Sprintf(`%s/v1/credentials/%s/status`, s.config.ServiceEndpoint, credID),
+			Type:                 statussdk.StatusList2021EntryType,
+			StatusPurpose:        statussdk.StatusRevocation,
+			StatusListIndex:      strconv.Itoa(statusListIndex),
+			StatusListCredential: statusListCredentialID,
+		}
+
+		if err := builder.SetCredentialStatus(status); err != nil {
+			errMsg := fmt.Sprintf("could not set credential status")
+			return nil, util.LoggingErrorMsg(err, errMsg)
```
decentralgabe, line 216:
```suggestion
			return nil, util.LoggingErrorMsg(err, "could not set credential status")
```

pkg/service/credential/service.go
decentralgabe: interesting, i can find these comments helpful but I see your point!

pkg/service/credential/service.go
```diff
@@ -338,6 +412,139 @@ func (s Service) GetCredentialsBySchema(request GetCredentialBySchemaRequest) (*
...
+	if err != nil {
+		return nil, util.LoggingNewErrorf("problem with getting status list credential for issuer: %s schema: %s", gotCred.Issuer, gotCred.Schema)
+	}
+
+	creds, err := s.storage.GetCredentialsByIssuerAndSchema(gotCred.Issuer, gotCred.Schema)
+	if err != nil {
+		return nil, util.LoggingNewErrorf("problem with getting status list credential for issuer: %s schema: %s", gotCred.Issuer, gotCred.Schema)
+	}
+
+	revokedStatusCreds := make([]credential.VerifiableCredential, 0)
```
decentralgabe, line 496:
```suggestion
	var revokedStatusCreds []credential.VerifiableCredential
```

pkg/service/credential/service.go
```diff
@@ -338,6 +412,139 @@ func (s Service) GetCredentialsBySchema(request GetCredentialBySchemaRequest) (*
...
+	return &response, nil
+}
+
+func (s Service) GetCredentialsByIssuerAndSchemaWithStatus(issuer string, schema string) ([]credential.VerifiableCredential, error) {
+	gotCreds, err := s.storage.GetCredentialsByIssuerAndSchema(issuer, schema)
+	if err != nil {
+		return nil, util.LoggingErrorMsgf(err, "could not get credential(s) for issuer: %s", issuer)
+	}
+
+	creds := make([]credential.VerifiableCredential, 0)
```
decentralgabe, line 538:
```suggestion
	var creds []credential.VerifiableCredential
```

pkg/service/credential/storage/bolt.go
decentralgabe: it's because the namespace bolt is using is created on first write. we should open up another issue to address it.
solutions:

1. seed the DB with a first entry 
2. detect the namespace not exists error and throw another error

pkg/service/credential/storage/bolt.go
```diff
@@ -1,35 +1,138 @@
...
+
 func NewBoltCredentialStorage(db *storage.BoltDB) (*BoltCredentialStorage, error) {
 	if db == nil {
 		return nil, errors.New("bolt db reference is nil")
 	}
+
+	// TODO: (Neal) there is a current bug with our Bolt implementation where if we do a GET without anything in the db it will throw an error
+	// Doing initial writes and then deleting will "warm up" our database and when we do a GET after that it will not crash and return empty list
+	err := db.Write(credentialNamespace, "fake-key", nil)
+	if err != nil {
```
decentralgabe, line 41:
```suggestion
	if err := db.Write(credentialNamespace, "fake-key", nil); err != nil {
```

pkg/service/credential/storage/bolt.go
```diff
@@ -1,35 +1,138 @@
...
+		return nil, util.LoggingErrorMsg(err, "problem writing status initial write to db")
+	}
+
+	err = db.Delete(credentialNamespace, "fake-key")
+	if err != nil {
```
decentralgabe, line 46:
```suggestion
	if err = db.Delete(credentialNamespace, "fake-key"); err != nil {
```

pkg/service/credential/storage/bolt.go
```diff
@@ -1,35 +1,138 @@
...
+		return nil, util.LoggingErrorMsg(err, "problem with initial delete to db")
+	}
+
+	err = db.Write(statusListCredentialNamespace, "fake-key", nil)
+	if err != nil {
```
decentralgabe, line 51:
```suggestion
	if err = db.Write(statusListCredentialNamespace, "fake-key", nil); err != nil {
```

pkg/service/credential/storage/bolt.go
```diff
@@ -1,35 +1,138 @@
...
+		return nil, util.LoggingErrorMsg(err, "problem writing status initial write to db")
+	}
+
+	err = db.Delete(statusListCredentialNamespace, "fake-key")
+	if err != nil {
```
decentralgabe, line 56:
```suggestion
	if err = db.Delete(statusListCredentialNamespace, "fake-key"); err != nil {
```

pkg/service/credential/storage/bolt.go
```diff
@@ -1,35 +1,138 @@
...
+		return nil, util.LoggingErrorMsg(err, "problem with generating random unique list")
+	}
+
+	uniqueNumBytes, err := json.Marshal(randUniqueList)
+	if err != nil {
+		return nil, util.LoggingErrorMsg(err, "could not marshal random unique numbers")
+	}
+
+	err = db.Write(statusListIndexNamespace, "status-list-indexes", uniqueNumBytes)
+	if err != nil {
```
decentralgabe, line 71:
```suggestion
	if err = db.Write(statusListIndexNamespace, "status-list-indexes", uniqueNumBytes); err != nil {
```

decentralgabe: let's make `"status-list-indexes"` a const

pkg/service/credential/storage/bolt.go
```diff
@@ -1,35 +1,138 @@
...
+		return nil, util.LoggingErrorMsg(err, "problem writing status list indexes to db")
+	}
+
+	statusListIndexBytes, err := json.Marshal(StatusListIndex{Index: 0})
+	if err != nil {
+		return nil, util.LoggingErrorMsg(err, "could not marshal status list index bytes")
+	}
+
+	err = db.Write(statusListIndexNamespace, "current-list-index", statusListIndexBytes)
+	if err != nil {
```
decentralgabe, line 81:
```suggestion
	if err = db.Write(statusListIndexNamespace, "current-list-index", statusListIndexBytes); err != nil {
```

decentralgabe: and `"current-list-index"`

pkg/service/credential/storage/bolt.go
decentralgabe: need to improve this everywhere

pkg/service/credential/storage/bolt.go
```diff
@@ -1,35 +1,138 @@
...
+		return -1, util.LoggingErrorMsgf(err, "could not unmarshal unique numbers")
+	}
+
+	statusListIndexBytes, err := json.Marshal(StatusListIndex{Index: statusListIndex.Index + 1})
+	if err != nil {
+		return -1, util.LoggingErrorMsg(err, "could not marshal status list index bytes")
+	}
+
+	err = b.db.Write(statusListIndexNamespace, "current-list-index", statusListIndexBytes)
+	if err != nil {
```
decentralgabe, line 120:
```suggestion
	if err = b.db.Write(statusListIndexNamespace, "current-list-index", statusListIndexBytes); err != nil {
```

thereisnogabe (changes requested):

pkg/service/credential/storage/storage.go
```diff
@@ -47,7 +48,15 @@ type Storage interface {
...
 	GetCredentialsBySchema(schema string) ([]StoredCredential, error)
+	GetCredentialsByIssuerAndSchema(issuer, schema string) ([]StoredCredential, error)
 	DeleteCredential(id string) error
+
+	StoreStatusListCredential(request StoreCredentialRequest) error
+	GetStatusListCredential(id string) (*StoredCredential, error)
+	GetStatusListCredentialsByIssuerAndSchema(issuer, schema string) ([]StoredCredential, error)
+	DeleteStatusListCredential(id string) error
+
+	GetNextStatusListRandomIndex() (int, error)
```
decentralgabe, line 59: wonder if this needs to be on the interface

pkg/server/router/credential.go
nitro-neal: I'll expand this status object to have revoked and suspended. 
In the future if there is a need to have a lot more status we can change it to be open eneded.

pkg/server/router/credential.go
nitro-neal: +2

pkg/service/credential/storage/bolt.go
nitro-neal: issue created - https://github.com/TBD54566975/ssi-service/issues/176

pkg/service/credential/storage/bolt.go
```diff
@@ -276,3 +448,29 @@ func (b BoltCredentialStorage) DeleteCredential(id string) error {
 func createPrefixKey(id, issuer, subject, schema string) string {
 	return strings.Join([]string{id, "is:" + issuer, "su:" + subject, "sc:" + schema}, "-")
 }
+
+func randomUniqueNum(count int) ([]int, error) {
```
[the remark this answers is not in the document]
nitro-neal: this is awesome, thanks!

config/compose.toml
nitro-neal: agree, I'll put it in top level config and propagate that down so every service has it available

internal/credential/model.go
nitro-neal: for now I'll add revoked and suspended and in the future if there is a need we can make it generic

pkg/server/router/credential.go
decentralgabe: sounds good perhaps a new issue that you can reference in code that covers this

internal/credential/model.go
nitro-neal: created issue -https://github.com/TBD54566975/ssi-service/issues/177

pkg/server/router/credential.go
nitro-neal: Added new issue that adds suspended in the future - https://github.com/TBD54566975/ssi-service/issues/177

pkg/server/router/credential_test.go
decentralgabe: I opened #173

thereisnogabe (approved):

PR #172, opened 2022-11-11
[OSE-167] Scaffolding for /v1/operations

andresuribe87:
This PR introduces the boilerplate needed to support the operation api with the get/list/cancel actions.


pkg/service/operation/service.go
```diff
@@ -0,0 +1,45 @@
+package operation
+
+import (
+	"fmt"
+	sdkutil "github.com/TBD54566975/ssi-sdk/util"
+	"github.com/pkg/errors"
+	"github.com/tbd54566975/ssi-service/internal/util"
+	"github.com/tbd54566975/ssi-service/pkg/service/framework"
+	storage2 "github.com/tbd54566975/ssi-service/pkg/service/operation/storage"
```
decentralgabe, line 9: perhaps `opstorage` as an alias instead of `storage2`?

pkg/service/operation/storage/bolt.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"github.com/pkg/errors"
+	"github.com/tbd54566975/ssi-service/pkg/storage"
+)
+
+type BoltOperationStorage struct {
+	db *storage.BoltDB
+}
+
+func (b BoltOperationStorage) StoreOperation(op StoredOperation) error {
```
decentralgabe, line 12: is this PR intended to be a draft? or do you aim to merge with these unimplemented?

andresuribe87: I intend to merge with these unimplemented. They will be implemented in subsequent PRs. Just trying to keep the surface area small.

pkg/service/operation/service.go
andresuribe87: Done.

decentralgabe (approved):
lgtm - need to fix a linter error before merge

PR #188, opened 2022-11-30
make storage more friendly

suutaku:
1. Add StorageParam to configure to support any addition storage configure.
2. Add `Init(interface{})` function to `ServiceStorage` interface.
3. Modify storage functions.
4. Register storage implementation at ` init()`


pkg/server/router/credential_test.go
```diff
@@ -41,7 +41,8 @@ func TestCredentialRouter(t *testing.T) {
 
 	t.Run("Credential Service Test", func(tt *testing.T) {
 
-		bolt, err := storage.NewBoltDB()
+		bolt := &storage.BoltDB{}
```
decentralgabe, line 44: why not `storage.NewBoltDB()`?

pkg/storage/bolt.go
```diff
@@ -25,17 +29,21 @@ func (b *BoltDB) Type() Storage {
...
 }
 
-// NewBoltDB instantiates a file-based storage instance for Bolt https://github.com/boltdb/bolt
-func NewBoltDB() (*BoltDB, error) {
-	return NewBoltDBWithFile(DBFile)
-}
-
-func NewBoltDBWithFile(filePath string) (*BoltDB, error) {
+// Init instantiates a file-based storage instance for Bolt https://github.com/boltdb/bolt
+func (b *BoltDB) Init(param interface{}) error {
```
decentralgabe, line 33: is `param` supposed to be generic for all storage methods?

if so...I think this needs to be variadic or use an options pattern

if not...let's make this `filePath string`

pkg/storage/bolt_test.go
```diff
@@ -1,14 +1,16 @@
...
 	"testing"
+
+	"github.com/goccy/go-json"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestBoltDB(t *testing.T) {
-	db, err := NewBoltDBWithFile("test.db")
+	db := &BoltDB{}
+	err := db.Init("test.db")
```
decentralgabe, line 13: I'm not sure I see the benefit in this type of pattern

i don't think users should be instantiating `&BoltDB{}` directly - a method would be preferable as there was before

pkg/storage/storage.go
```diff
@@ -21,28 +24,31 @@ type ServiceStorage interface {
...
+	impl := GetStorage(storageProvider)
+	if impl == nil {
+		return impl, fmt.Errorf("unsupported storage provider: %s", storageProvider)
 	}
+	err := impl.Init(param)
+	return impl, err
 }
 
-// AvailableStorage returns the supported storage providers
-func AvailableStorage() []Storage {
```
decentralgabe, line 35: I think this method is still useful for "default" storage providers

pkg/server/router/credential_test.go
suutaku: For dynamic register，implementations should load to `avaliableStorages` with `init()` in  implementation file `xxx_storage.go`.  this `xxx_storage.go` can be a part of `ssi-service` or a 3rd party repo which use `ssi-service` as a library.
if we use functions like `NewBoltDB `, different implementations will use different `NewXXXX` function so we can not create and set up implementations in `NewStorage` automaticlly (i also add an uniform standard `Init(interface{})` into ServiceStorage, so we can do `set up` with params for implementation ).

pkg/storage/bolt.go
suutaku: yes, `param`  supposed to be generic for all storage methods.  the problem is we don't know what param should be. For a kv-store, it can be a file path, but for blockchain or something else, it can be a struct.

pkg/storage/bolt_test.go
suutaku: let's see the code blew:

```go
// NewStorage creates a new storage provider based on the input
func NewStorage(storageProvider Storage) (ServiceStorage, error) {
	switch storageProvider {
	case Bolt:
		return NewBoltDB()
	default:
		return nil, fmt.Errorf("unsupported storage provider: %s", storageProvider)
	}
}
```
think about if we  don't use current pattern.  every time we add an implementation, we should add one case to `NewStorage`. and this way also not support 3rd party register.

but, use current pattern, Whatever new implementation added or  3rd party implementation registed, `NewStorage` code will not change.  just like blew:

```go
// NewStorage creates a new storage provider and provider specify param based on the input
func NewStorage(storageProvider Storage, param interface{}) (ServiceStorage, error) {
	impl := GetStorage(storageProvider)
	if impl == nil {
		return impl, fmt.Errorf("unsupported storage provider: %s", storageProvider)
	}
	err := impl.Init(param)
	return impl, err
}
```

>> i don't think users should be instantiating &BoltDB{} directly - a method would be preferable as there was before

if we use functions like NewBoltDB , different implementations will use different NewXXXX function so we can not create and set up implementations in NewStorage automaticlly (i also add an uniform standard Init(interface{}) into ServiceStorage, so we can do set up with params for implementation ).

```go
...
err := impl.Init(param)
...
```

pkg/storage/storage.go
suutaku: if we have "default"  storage setting in configure, this will not be a problem. but i think will still need `AvailableStorage ` for example, show a list of available storages.  so keep it ?

pkg/storage/bolt.go
suutaku: and yes, it can be variadic or use an options pattern. but in `Init()`, we always should input a value to function, a implementation specify struct (or type), or just a `nil`. 
i have thinking through this kind of situation:
```go
func (b *BoltDB) Init(params ...interface{}) error
func (b *BoltDB) Init(params ...string) error
```
but i think `func (b *BoltDB) Init(param interface{}) error` was the best solution.
if you have any idea, please tell me. we can talk about it deeply.

suutaku: > if so...I think this needs to be variadic or use an options pattern
> 
by the way, i really should change `param` to `options`.

pkg/storage/bolt_test.go
decentralgabe: Thanks @suutaku I think this is the key of my gap in understanding.

We can make the pattern more generic as you have here, but practically - how will a 3rd party option become registered?

I don't foresee anyone using this project as a library. To modify it they would need to fork and make changes. With that burden, does this change really make much simpler?

From another perspective I can see not wanting to have any DB be the "default" for the service. I think that is a valuable change (though may already be possible?).

decentralgabe:
curious what @andresuribe87 thinks about this pattern

pkg/storage/bolt_test.go
andresuribe87: I have the same question as Gabe. When adding a new storage, is it simpler to add an import with side effects, or to add a new case in the switch statement? 

Additionally, related to `init` functions. How would we enforce that developers do registration of the storage within the module (or package) they are adding?

suutaku: @decentralgabe @andresuribe87 Thanks to replay me! About questions:
> The case of use `ssi-service` as a library

 `ssi-service` can be a part of another service,  currently, we can use `micro-service way` to do that, but i think we also should support `import way`.
A simple example, i have a repo, named `my-repo`, i have a new implementation of storage (can be a 3rd party repo, or just a single .go file in `my-repo`).  at some where of my code, I import `ssi-service`, called `RegisterStorage`, use `ssi-service` as a instance and everything is done.

> how will a 3rd party option become registered?

you mean options in configure ?  When developers call `RegisterStorage` register 3rd party storage, option is not registered.  developers should modify configure file (the `StorageParam` or `StorageOptions` tag) and the `NewStorage` called, storage specify 3rd party options will applied according to `Init(interface{})` function's implementation. in `Init(interface{})` function developer can do everything they want to do with the option(s).

>  How would we enforce that developers do registration of the storage within the module (or package) they are adding?

We cannot force developer do anything. we can just tell developers, there was a simple way to use our code: import it, call `RegiserStorage` and Ok.  in fact,  they can call `RegiserStorage` any place the want, the `init()` function just an option.

Note that, in this approach, developers will not modify any single code in `ssi-service`.

suutaku: @decentralgabe I agree with you. the "default" (currently it is `bolt`) just make `ssi-service` working.  
> From another perspective I can see not wanting to have any DB be the "default" for the service. I think that is a valuable change (though may already be possible?).

andresuribe87: > at some where of my code, I import ssi-service, called RegisterStorage, use ssi-service as a instance and everything is done.

Gotcha! I now understand better what you're trying to accomplish, and the registry pattern certainly does it well. 

That said, I would prefer explicit dependency injection so developers have an easier time tracking down what instances are being used. In this case, I propose the following alternative. 

1. Modify the `ServicesConfig` struct to have a method called `ServiceStorage`. 
2. The `ServiceStorage` would have a default implementation that returns the bold storage.

Devs can then inject whatever `ServiceStorage` implementation they like, without the need of modifying the underlying code. 

The benefits of this approach is that we don't use the `init` mechanism, which I've seen in the past is difficult to understand.

suutaku: yes, that's also a good solution! in face, we can also don't use the init mechanism, just call `RegisterStorage` some where. 
and, I would prefer  keep the `availableStorage map[Storage]StorageService`, because with it, we can event switch storage in dynamic(the case of use multiple storage).
so the flow can be: setting multiple storage string in configure file ->  inject multiple storage to service -> make a switch logic -> use one of a storage.

pkg/storage/storage.go
decentralgabe: yes

pkg/storage/bolt_test.go
decentralgabe: agree with this approach!

pkg/server/router/credential_test.go
decentralgabe: in general I find side-effects like "init" dangerous because their behaviors are not immediately clear, and it can be confusing to debug. if we can maintain the behavior of init but make it explicit that would be preferable.

I think we can create an interface `NewStorage` which `NewBoltDB` and others would comply with to mitigate your concern

pkg/storage/bolt_test.go
suutaku: Let's make `StorageService` to manage `Storage` implementations first? Use `ServicesConfig` do that will be a little confused? in my opinion, the configure only manage the configure files, it' s text work. 
also some day, we need configure watching the conf files if it was modified,  that's one of the way we can change (switch) settings dynamically.

pkg/service/credential/storage.go
```diff
@@ -225,7 +262,7 @@ func (b BoltCredentialStorage) getCredential(id string, namespace string) (*Stor
 // GetCredentialsByIssuer gets all credentials stored with a prefix key containing the issuer value
 // The method is greedy, meaning if multiple values are found and some fail during processing, we will
 // return only the successful values and log an error for the failures.
-func (b BoltCredentialStorage) GetCredentialsByIssuer(issuer string) ([]StoredCredential, error) {
+func (b *CredentialStorage) GetCredentialsByIssuer(issuer string) ([]StoredCredential, error) {
 	keys, err := b.db.ReadAllKeys(credentialNamespace)
```
decentralgabe, line 266: I really like this change but I'm curious how feasible it is to map database behavior across impls. it's worth giving it a try, I think!

pkg/service/keystore/storage.go
```diff
@@ -39,7 +62,7 @@ func NewBoltKeyStoreStorage(db *storage.BoltDB, key ServiceKey) (*BoltKeyStoreSt
 }
 
 // TODO(gabe): support more robust service key operations, including rotation, and caching
-func (b *BoltKeyStoreStorage) storeServiceKey(key ServiceKey) error {
+func (b *KeyStoreStorage) storeServiceKey(key ServiceKey) error {
```
decentralgabe, line 65: nit: `s` or `ks` or `kss`

pkg/service/presentation/storage.go
```diff
@@ -16,11 +18,42 @@ const (
...
+}
+
+var ErrSubmissionNotFound = errors.New("submission not found")
+
+type PresentationStorage struct {
+	db storage.ServiceStorage
 }
 
-func (b BoltPresentationStorage) ListSubmissions() ([]StoredSubmission, error) {
+func (b *PresentationStorage) ListSubmissions() ([]StoredSubmission, error) {
```
decentralgabe, line 56: nit: `s` or `ps`

pkg/storage/bolt.go
```diff
@@ -14,9 +14,14 @@ import (
 )
 
 const (
-	DBFile = "ssi-service.db"
+	Bolt Storage = "bolt"
 )
 
+// try to register your implementation at your .go file like this:
+func init() {
+	RegisterStorage(&BoltDB{})
```
decentralgabe, line 22: nit: `RegisterStorage(new(BoltDB))`

pkg/storage/bolt.go
```diff
@@ -25,17 +30,35 @@ func (b *BoltDB) Type() Storage {
...
-		return nil, err
+		return err
 	}
-	return &BoltDB{db: db}, nil
+	b.db = db
+	return nil
+}
+
+// Uri return filepath of boltDB,
+func (b *BoltDB) Uri() string {
```
decentralgabe, line 60:
```suggestion
func (b *BoltDB) FilePath() string {
```

pkg/storage/bolt_test.go
```diff
@@ -94,7 +96,8 @@ func TestBoltDB(t *testing.T) {
 }
 
 func TestBoltDBPrefixAndKeys(t *testing.T) {
-	db, err := NewBoltDBWithFile("test.db")
+	db := &BoltDB{}
```
decentralgabe, line 99:
```suggestion
	db := new(BoltDB)
```

pkg/storage/storage.go
```diff
@@ -1,48 +1,70 @@
...
 
 type Storage string
 
 const (
-	Bolt Storage = "bolt"
+	DBFilePrefix = "ssi-service"
+)
+
+var (
+	availableStorages = make(map[Storage]ServiceStorage)
```
decentralgabe, line 12: nit: I would prefer to initialize this in a `NewServiceStorage` method or similar

decentralgabe (commented):
looks great, few small comments.

michaelneale:
this is super cool BTW

pkg/storage/storage.go
suutaku: Its purpose is automatic run initialize. i put this initialization to the `init()` before, but we cannot guarantee this `init()` was the first execution (another `init()` was in the `bolt.go` ). That's why, in the end, I've made a choice.
if we create a function like `NewServiceStorage `, we should consider that when and where we call it, things back to the beginning. so think this approach was not necessary.

pkg/service/presentation/storage.go
decentralgabe: one more here

pkg/storage/bolt.go
decentralgabe: maybe `ResourcePath`? 

fine to keep it as Uri, but please make it `URI`

pkg/storage/bolt.go
```diff
@@ -14,9 +14,14 @@ import (
...
 
 const (
-	DBFile = "ssi-service.db"
+	Bolt Storage = "bolt"
 )
 
+// try to register your implementation at your .go file like this:
+func init() {
+	RegisterStorage(new(BoltDB))
+}
```
decentralgabe, line 23: the problem I have with this is - if I am running this service in production, and want to use another storage provider...do I now have multiple storages running at the same time?

i could see a world where there are multiple storage providers...but as written now the code assumes one. does introduce a path where that constraint is broken?

pkg/storage/bolt.go
```diff
@@ -14,9 +14,14 @@ import (
 )
 
 const (
-	DBFile = "ssi-service.db"
+	Bolt Storage = "bolt"
```
decentralgabe, line 17:
```suggestion
	BoltStorage = "bolt"
```

pkg/storage/bolt.go
suutaku: `RegisterStorage(new(BoltDB))` just make a pointer of  bolt storage. like write a selection on paper. the real process who make and initialize the bolt storage is `storage.Init(interface{})` at `NewStorage()`. 
so after `RegisterStorage ` done, we still have no useable storage instance. At this time we only have a **canability* *that we can use `NewStorage` to choice a storage and make it useable.
For example, `ssi-sevice` have a `default` storage selection, it's `Bolt`. Developer add  `LevelDB` and `Badger` to service with `RegisterStorage`. now he can select use `Bolt`, `LevelDB` or `Badger` with calling `NewStorage`, if he selected `Badger`, `NewStorage` get `Badger` from the`valiableStorage`, create or open db file in `storage.Init(interface{})`. now `Badger` is ready to use.

suutaku (commented):
review done

decentralgabe:
looks like a number of conflicts. happy to merge once they're resolved

pkg/storage/bolt.go
decentralgabe: thanks this makes sense

decentralgabe (approved):

suutaku:
this pull request change source structures and missing new features,  conflicts too complex to resolve, should I close it and re-fork?

decentralgabe:
sure, up to you

PR #194, opened 2022-12-01
[OSE-175] GetOperations implementation

andresuribe87:
# Overview
GetOperations implementation. Notably, this doesn't include any filtering logic, which will be added in a subsequent PR. 

# Description

# How Has This Been Tested?
Unit test. 

# Checklist

## References
See SIP6.

### DIFF
--- a/pkg/server/server_operation_test.go
+++ b/pkg/server/server_operation_test.go
@@ -1,0 +1,84 @@
+package server
+
+import (
+	"github.com/TBD54566975/ssi-sdk/credential/exchange"
+	"github.com/goccy/go-json"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/stretchr/testify/assert"
+	"github.com/tbd54566975/ssi-service/pkg/server/router"
+	"github.com/tbd54566975/ssi-service/pkg/service/operation"
+	"github.com/tbd54566975/ssi-service/pkg/storage"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"testing"
+)
+
+func TestOperationsAPI(t *testing.T) {
+	t.Run("GetOperations", func(t *testing.T) {
+		t.Run("Returns empty when no operations stored", func(t *testing.T) {
+			opRouter := setupOperationsRouter(t, nil)
+
+			request := router.GetOperationsRequest{}
+			value := newRequestValue(t, request)
+			req := httptest.NewRequest(http.MethodGet, "https://ssi-service.com/v1/operations", value)
+			w := httptest.NewRecorder()
+
+			assert.NoError(t, opRouter.GetOperations(newRequestContext(), w, req))
+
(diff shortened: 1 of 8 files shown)


pkg/server/server_operation_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/tbd54566975/ssi-service/pkg/storage"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"testing"
+)
+
+func TestOperationsAPI(t *testing.T) {
+	t.Run("GetOperations", func(t *testing.T) {
+		t.Run("Returns empty when no operations stored", func(t *testing.T) {
```
decentralgabe, line 20: nit: on the inner runners you can use different variables than `t` to not shadow the parent t's

pkg/server/server_operation_test.go
```diff
@@ -0,0 +1,83 @@
...
+			value := newRequestValue(t, request)
+			req := httptest.NewRequest(http.MethodGet, "https://ssi-service.com/v1/operations", value)
+			w := httptest.NewRecorder()
+
+			assert.NoError(t, opRouter.GetOperations(newRequestContext(), w, req))
+
+			var resp router.GetOperationsResponse
+			assert.NoError(t, json.NewDecoder(w.Body).Decode(&resp))
+			assert.Empty(t, resp.Operations)
+		})
```
decentralgabe, line 35: newline for readability between these plz

pkg/service/operation/service.go
```diff
@@ -32,6 +32,30 @@ func (s Service) Status() framework.Status {
 	return framework.Status{Status: framework.StatusReady}
 }
 
+func (s Service) GetOperations(request GetOperationsRequest) (*GetOperationsResponse, error) {
+	ops, err := s.storage.GetOperations()
```
decentralgabe, line 36: could consider a cache on `s Service` in the future

decentralgabe (approved):
🥳

pkg/service/operation/service.go
andresuribe87: yes, definitely something we should think about

pkg/server/server_operation_test.go
andresuribe87: I'm of the opinion that shadowing makes sense here, unless there is a strong reason not to do so. If you find yourself having to access the outer `t` from inside a subtest, that's an indication that there is something unusual happening. Curious what your thoughts are?

pkg/server/server_operation_test.go
andresuribe87: Done

pkg/server/server_operation_test.go
decentralgabe: have run into shadowing errs with unexpected state, especially when tests are run in parallel. IMO as much as possible we should make our tests be able to run in parallel, which means eliminating shared state. The test as written is fine - so no problem.

PR #195, opened 2022-12-02
Add did web integ test

nitro-neal:
Adding new did:web test.

This test runs the steelthread workflow but with a did:web instead.


magefile.go
```diff
@@ -115,10 +115,25 @@ func Spec() error {
 func Integration(testName string) error {
 	switch testName {
 	case "steelthread":
-		return steelthread.RunTest()
-	default:
+		return integtest.RunSteelthreadTest()
+	case "didweb":
+		return integtest.RunDidWebTest()
```
decentralgabe, line 120: nit: `RunDIDWebTest`

magefile.go
```diff
@@ -115,10 +115,25 @@ func Spec() error {
...
+	case "all":
 		fmt.Println("Running all tests")
-		return steelthread.RunTest()
+		err := integtest.RunSteelthreadTest()
+		if err != nil {
```
decentralgabe, line 124:
```suggestion
		if err := integtest.RunSteelthreadTest(); err != nil {
```

magefile.go
```diff
@@ -115,10 +115,25 @@ func Spec() error {
...
+			fmt.Printf("Error running steelthread test")
+			return err
+		}
+		err = integtest.RunDidWebTest()
+		if err != nil {
```
decentralgabe, line 129:
```suggestion
		if err = integtest.RunDIDWebTest(); err != nil {
```

magefile.go
```diff
@@ -115,10 +115,25 @@ func Spec() error {
...
+			fmt.Printf("Error running did:web steelthread test")
```
decentralgabe, line 130: I don't remember are we using logrus for the rest of this file? I think print is fine

test/common.go
```diff
@@ -0,0 +1,131 @@
...
+	"embed"
+	cmpact "encoding/json"
+	"github.com/goccy/go-json"
+	"fmt"
+	manifestsdk "github.com/TBD54566975/ssi-sdk/credential/manifest"
+	"github.com/oliveagle/jsonpath"
+	"github.com/pkg/errors"
+	credmodel "github.com/tbd54566975/ssi-service/internal/credential"
+	"io"
+	"net/http"
```
decentralgabe, line 14: might have to run [goimports](https://pkg.go.dev/golang.org/x/tools/cmd/goimports) here

test/common.go
```diff
@@ -0,0 +1,131 @@
...
+	jsonBytes := []byte(json)
+	buffer := new(bytes.Buffer)
+	if err := cmpact.Compact(buffer, jsonBytes); err != nil {
+		fmt.Println(err)
+	}
+
+	return buffer.String()
+}
+
+func getJSONElement(jsonString string, jsonPath string) (string, error) {
```
decentralgabe, line 37: this is a great function, could move it into the `util` package

thereisnogabe (commented):
the new `test` directory - should it be named something like `integration`?

magefile.go
```diff
@@ -115,10 +115,25 @@ func Spec() error {
 func Integration(testName string) error {
 	switch testName {
 	case "steelthread":
-		return steelthread.RunTest()
-	default:
+		return integtest.RunSteelthreadTest()
```
decentralgabe, line 118: nit: `RunSteelThreadTest`

test/didweb.go
```diff
@@ -0,0 +1,144 @@
...
+	"fmt"
+	"github.com/TBD54566975/ssi-sdk/crypto"
+	"github.com/mr-tron/base58"
+	"github.com/tbd54566975/ssi-service/internal/keyaccess"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+func RunDidWebTest() error {
```
decentralgabe, line 13: I think all tests should have the `_test.go` suffix and we should use the `testing.T` paradigm.

For integration tests, let's take advantage of the [paradigm outlined here](https://stackoverflow.com/a/41407042) so we can selectively run unit/integration tests in our pipelines.

magefile.go
```diff
@@ -115,10 +115,25 @@ func Spec() error {
 func Integration(testName string) error {
```
decentralgabe, line 115: After [this comment](https://github.com/TBD54566975/ssi-service/pull/195/files#r1038413605) this method will change to use the go test runner with an `integration` flag.

test/didweb.go
decentralgabe, line 13:
```suggestion
func RunDIDWebTest() error {
```

test/didweb.go
```diff
@@ -0,0 +1,144 @@
...
+	if err != nil {
+		return errors.Wrap(err, "problem getting json element")
+	}
+	manifestID, err := getJSONElement(output, "$.credential_manifest.id")
+	if err != nil {
+		return errors.Wrap(err, "problem getting json element")
+	}
+
+	// Submit an application
+	fmt.Println("\n\nSubmit an Application:")
```
decentralgabe, line 99: similarly all Print statements should be replaced with `t.Log`...

test/didweb.go
```diff
@@ -0,0 +1,144 @@
...
+	applicationJSON := getJSONFromFile("application-input.json")
+	applicationJSON = strings.ReplaceAll(applicationJSON, "<DEFINITIONID>", presentationDefinitionID)
+	applicationJSON = strings.ReplaceAll(applicationJSON, "<VCJWT>", credentialJWT)
+	applicationJSON = strings.ReplaceAll(applicationJSON, "<MANIFESTID>", manifestID)
+
+	// Start signing credential application
+	alicPrivKeyBytes, err := base58.Decode(aliceDIDPrivateKey)
```
decentralgabe, line 106: alice!

thereisnogabe (changes requested):
see comments - I know this is a deviation from the use cases in the ssi-sdk and other tests of this style. those should be fixed too 😄

test/common.go
andresuribe87: would suggest adding `goimports` to the golang-ci file :)

test/common.go
```diff
@@ -0,0 +1,131 @@
...
+var (
+	//go:embed testdata
+	testVectors embed.FS
+)
+
+func compactJSONOutput(json string) string {
+	jsonBytes := []byte(json)
+	buffer := new(bytes.Buffer)
+	if err := cmpact.Compact(buffer, jsonBytes); err != nil {
+		fmt.Println(err)
```
andresuribe87, line 31: For test utils, a good pattern is to simply panic here. An error log statement be easily missed.

test/common.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	return buffer.String()
+}
+
+func getJSONElement(jsonString string, jsonPath string) (string, error) {
+	jsonMap := make(map[string]interface{})
+	if err := json.Unmarshal([]byte(jsonString), &jsonMap); err != nil {
+		return "", errors.Wrap(err, "problem with unmarshalling json string")
```
andresuribe87, line 40: Errors are typically already indicative of problems, so removing it would preserve the semantics.
```suggestion
		return "", errors.Wrap(err, "unmarshalling json string")
```

test/common.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	element, err := jsonpath.JsonPathLookup(jsonMap, jsonPath)
+	if err != nil {
+		return "", errors.Wrap(err, "problem with finding element in json string")
```
andresuribe87, line 45:
```suggestion
		return "", errors.Wrap(err, "finding element in json string")
```

test/common.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	return element.(string), err
```
andresuribe87, line 48: For clarity
```suggestion
	return element.(string), nil
```

test/common.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func get(url string) (string, error) {
+	fmt.Printf("\nPerforming GET request to:  %s\n", url)
+
+	resp, err := http.Get(url) // #nosec: testing only.
+	if err != nil {
+		return "", errors.Wrap(err, "problem with finding element in json string")
```
andresuribe87, line 56:
```suggestion
		return "", errors.Wrap(err, fmt.Sprintf("GET-ing url <%s>", url))
```

test/common.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return "", errors.Wrap(err, "problem with parsing body")
```
andresuribe87, line 61: Here' and elsewhere. When wrapping errors, the context read's better when you describe what operation lead to the error.
```suggestion
		return "", errors.Wrap(err, "reading all body")
```

test/common.go
```diff
@@ -0,0 +1,131 @@
...
+	endpoint = "http://localhost:8080/"
+	version  = "v1/"
+)
+
+var (
+	//go:embed testdata
+	testVectors embed.FS
+)
+
+func compactJSONOutput(json string) string {
```
andresuribe87, line 27: Curious what's the reason for using the `encoding/json` vs the other json function?

test/common.go
```diff
@@ -0,0 +1,131 @@
...
+	jsonBytes := []byte(json)
+	buffer := new(bytes.Buffer)
+	if err := cmpact.Compact(buffer, jsonBytes); err != nil {
```
andresuribe87, line 30: Any particular reason to use `Compact`?

test/common.go
```diff
@@ -0,0 +1,131 @@
...
+	return string(body), err
+}
+
+func getJSONFromFile(fileName string) string {
+	b, _ := testVectors.ReadFile("testdata/" + fileName)
+	return string(b)
+}
+
+func is200Response(statusCode int) bool {
+	return statusCode/100 != 2
```
andresuribe87, line 111:
```suggestion
	return statusCode != 200
```

test/common.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func getValidApplicationRequest(credAppJSON string, credentialJWT string) manifestsdk.CredentialApplicationWrapper {
+	var createApplication manifestsdk.CredentialApplication
+	if err := json.Unmarshal([]byte(credAppJSON), &createApplication); err != nil {
+		fmt.Println("unmarshal error")
```
andresuribe87, line 117: Panic here... or better yet, include the `testing.T` variable and run assertions (or call `require.XXX`.

test/didweb.go
andresuribe87: Heavy +1 to this.

test/common.go
decentralgabe: good call - we should be using `go-json` everywhere

test/common.go
decentralgabe: believe this is fine as is - want to capture any 200 code

andresuribe87: ohhhhhh then I'd suggest renaming to `is2xxResponse`

magefile.go
nitro-neal: I use logrus everywhere now

test/didweb.go
nitro-neal: did this in the refactor

test/common.go
nitro-neal: it makes the massively large json output smaller so it doesn't take up the entire console window, so I compact

integration/common.go
```diff
@@ -1,22 +1,22 @@
-package steelthread
+package integration
 
 import (
 	"bytes"
 	"embed"
+	"encoding/json"
+	cmpact "encoding/json"
```
decentralgabe, line 7: we should be using `"github.com/goccy/go-json"` as a drop in replacement for encoding/json everywhere

integration/didweb_integration_test.go
```diff
@@ -0,0 +1,153 @@
+package integration
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+var didWebContext = NewTestContext("DidWeb")
```
decentralgabe, line 8: interesting...I'm hesitant of having global state like this. there could be side effects between tests which could produce some ugly errors.

what do you think about instantiating a new context per test? also nit `DIDWeb`

thereisnogabe (commented):
⭐ 👏 🎺  this is AWESOME! nice job @nitro-neal

integration/common.go
```diff
@@ -0,0 +1,245 @@
+package integration
+
+import (
+	"bytes"
+	"embed"
+	cmpact "encoding/json"
```
decentralgabe, line 6: still can remove this 😄

integration/didweb_integration_test.go
nitro-neal: I need to have shared state for the tests as some inputs are needed from previous tests.

I added a mutex around this and since each test is in it's own context there should be no interference betwen tests

integration/common.go
nitro-neal: my golang goes crazy and auto uses "encoding/json" even when I manually put in github.com/goccy/go-json,

it drives me crazy

decentralgabe: yeah it is super annoying. you could add a file watcher to auto replace https://www.jetbrains.com/help/go/settings-tools-file-watchers.html

nitro-neal:
❯ mage test
[test -tags=jwx_es256k -race -short ./...]?   	github.com/tbd54566975/ssi-service/cmd	[no test files]
ok  	github.com/tbd54566975/ssi-service/config	(cached)
ok  	github.com/tbd54566975/ssi-service/integration	0.173s
?   	github.com/tbd54566975/ssi-service/internal/credential	[no test files]
ok  	github.com/tbd54566975/ssi-service/internal/did	(cached)
ok  	github.com/tbd54566975/ssi-service/internal/keyaccess	(cached)
?   	github.com/tbd54566975/ssi-service/internal/schema	[no test files]
ok  	github.com/tbd54566975/ssi-service/internal/util	(cached)
?   	github.com/tbd54566975/ssi-service/pkg/jwt	[no test files]
ok  	github.com/tbd54566975/ssi-service/pkg/server	135.200s
?   	github.com/tbd54566975/ssi-service/pkg/server/framework	[no test files]
?   	github.com/tbd54566975/ssi-service/pkg/server/middleware	[no test files]
ok  	github.com/tbd54566975/ssi-service/pkg/server/router	12.859s
?   	github.com/tbd54566975/ssi-service/pkg/service	[no test files]
?   	github.com/tbd54566975/ssi-service/pkg/service/credential	[no test files]
?   	github.com/tbd54566975/ssi-service/pkg/service/credential/storage	[no test files]
?   	github.com/tbd54566975/ssi-service/pkg/service/did	[no test files]
?   	github.com/tbd54566975/ssi-service/pkg/service/did/storage	[no test files]
?   	github.com/tbd54566975/ssi-service/pkg/service/framework	[no test files]
ok  	github.com/tbd54566975/ssi-service/pkg/service/keystore	1.641s
?   	github.com/tbd54566975/ssi-service/pkg/service/keystore/storage	[no test files]
?   	github.com/tbd54566975/ssi-service/pkg/service/manifest	[no test files]
?   	github.com/tbd54566975/ssi-service/pkg/service/manifest/storage	[no test files]
ok  	github.com/tbd54566975/ssi-service/pkg/service/operation	0.392s
?   	github.com/tbd54566975/ssi-service/pkg/service/operation/storage	[no test files]
?   	github.com/tbd54566975/ssi-service/pkg/service/presentation	[no test files]
?   	github.com/tbd54566975/ssi-service/pkg/service/presentation/storage	[no test files]
?   	github.com/tbd54566975/ssi-service/pkg/service/schema	[no test files]
?   	github.com/tbd54566975/ssi-service/pkg/service/schema/storage	[no test files]
ok  	github.com/tbd54566975/ssi-service/pkg/storage	0.443s

❯ mage integration
[test -tags=jwx_es256k -race ./integration]ok  	github.com/tbd54566975/ssi-service/integration

This change does a few things:
Full refactor of our integration tests, now it uses the golang testing framework and is run with 'mage integration'
Fixed a bug in cred revocation with status not updating properly
Steelthread integration test rewrite
Did:web full integration test
Revocation full integration test

thereisnogabe (approved):

PR #197, opened 2022-12-06
[OSE-173] Filter implementation with tests

andresuribe87:
# Overview
This change adds filtering capabilities for the `GetOperations` and `ListSubmissions` endpoints. 

# Description
This implementation sticks to the following pattern:
* Parse the `filter` parameter in the request using the `einride` library. This parses into an `expression` protobuf.
* Evaluate the parsed expression by passing the variable values based off of what element was retrieved from storage.

The implementation relies on knowledge of google's CEL (see this [codelab](https://codelabs.developers.google.com/codelabs/cel-go/index.html#0) for a good intro). There are alternatives to doing this, but CEL makes it very easy to extend and to translate into multiple backends once we want to support more complex filtering queries quickly.

# How Has This Been Tested?
See unit tests added.

# Checklist

## References
* https://pkg.go.dev/go.einride.tech/aip@v0.60.0
* https://codelabs.developers.google.com/codelabs/cel-go/index.html#0


pkg/server/router/operation.go
```diff
@@ -47,15 +48,40 @@ type GetOperationsRequest struct {
...
+}
+
+func (r GetOperationsRequest) ToServiceRequest() (operation.GetOperationsRequest, error) {
+	var opReq operation.GetOperationsRequest
+
+	declarations, err := filtering.NewDeclarations(
+		filtering.DeclareFunction(filtering.FunctionEquals,
+			filtering.NewFunctionOverload(
+				filtering.FunctionOverloadEqualsString, filtering.TypeBool, filtering.TypeBool, filtering.TypeBool)),
+		filtering.DeclareIdent("done", filtering.TypeBool),
```
decentralgabe, line 66: are these the only filter types? done, true, false?

if so - may be good to define them as consts here

pkg/service/presentation/storage/bolt.go
```diff
@@ -20,19 +20,26 @@ type BoltPresentationStorage struct {
...
+	shouldInclude, err := storage.Evaluator(filter)
+	if err != nil {
+		return nil, err
+	}
+	storedSubmissions := make([]StoredSubmission, 0, len(allData))
 	for key, data := range allData {
-		if err := json.Unmarshal(data, &storedSubmissions[i]); err != nil {
-			logrus.WithError(err).WithField("key", key).Errorf("unmarshalling submission")
+		var ss StoredSubmission
+		if err := json.Unmarshal(data, &ss); err != nil {
```
decentralgabe, line 36: err is shadowed so can `err = json...`

pkg/service/presentation/storage/bolt.go
```diff
@@ -20,19 +20,26 @@ type BoltPresentationStorage struct {
...
+			logrus.WithError(err).WithField("key", key).Error("unmarshalling submission")
+		}
+		if !shouldInclude(ss) {
+			continue
 		}
-		i++
+		storedSubmissions = append(storedSubmissions, ss)
```
decentralgabe, line 42:
```go
if shouldInclude(ss) {
    storedSubmissions = append(storedSubmissions, ss)
}
```

pkg/storage/filter.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/common/types/ref"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"go.einride.tech/aip/filtering"
+)
+
+type foo interface {
```
decentralgabe, line 11: why foo?

pkg/storage/filter.go
```diff
@@ -0,0 +1,61 @@
...
+			logrus.WithError(err).
+				WithField("details", det).
+				Error("evaluating submission")
+			panic(err)
+		}
+		return out.Value() == true
+	}, nil
+}
+
+func Env() (*cel.Env, error) {
```
decentralgabe, line 46: please add comments to the methods in this file

decentralgabe (changes requested):

pkg/service/operation/storage/bolt.go
```diff
@@ -43,18 +44,31 @@ func (b BoltOperationStorage) GetOperation(id string) (*StoredOperation, error)
...
-		stored = append(stored, nextOp)
+		include, err := shouldInclude(nextOp)
+		// We explicitly ignore evaluation errors and simply include them in the result.
+		if err != nil {
+			stored = append(stored, nextOp)
+			continue
+		}
+		if include {
+			stored = append(stored, nextOp)
+		}
```
decentralgabe, line 71: could this be changed to something like this?

```suggestion
		if err != nil {
			logrus.WithErr(err).Error(...)
		}
		stored = append(stored, nextOp)
```

pkg/server/router/operation.go
```diff
@@ -47,15 +48,48 @@ type GetOperationsRequest struct {
...
 
-func (r GetOperationsRequest) ToServiceRequest() operation.GetOperationsRequest {
-	return operation.GetOperationsRequest{
-		Parent: r.Parent,
-		Filter: r.Filter,
+func (r GetOperationsRequest) GetFilter() string {
+	return r.Filter
+}
+
+const (
```
decentralgabe, line 59: nit: could define a type for these e.g. `type OperationsFilter string` however the type conversion may be a pain

andresuribe87: I don't expect these to be used elsewhere so leaving as is makes sense to me for now.

pkg/service/operation/storage/bolt.go
andresuribe87: Errors are already logged inside the `NewIncludeFunc`

Separately, your suggestion doesn't check the `include` variable, which would break the filtering behavior.

pkg/service/presentation/storage/bolt.go
andresuribe87: Ack.  Used the previously declared var.

pkg/service/presentation/storage/bolt.go
```diff
@@ -20,19 +20,31 @@ type BoltPresentationStorage struct {
...
 	for key, data := range allData {
-		if err := json.Unmarshal(data, &storedSubmissions[i]); err != nil {
-			logrus.WithError(err).WithField("key", key).Errorf("unmarshalling submission")
+		var ss StoredSubmission
+		if err := json.Unmarshal(data, &ss); err != nil {
+			logrus.WithError(err).WithField("key", key).Error("unmarshalling submission")
+		}
+		include, err := shouldInclude(ss)
+		// We explicitly ignore evaluation errors and simply include them in the result.
+		if err != nil {
```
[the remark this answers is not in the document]
andresuribe87: See above.

pkg/service/operation/storage/bolt.go
decentralgabe: how does it break the behavior? as I read it your code does the same behavior whether or not the include var is checked

decentralgabe (approved):

pkg/service/operation/storage/bolt.go
andresuribe87: Suppose that the `shouldInclude` function returns:
 * `include` equal to `false`
 * `err` equal to `nil` 

Under the above scenario, your suggestion would include `nextOp`; the current implementation would not.

decentralgabe: ah, gotcha. guess it's fine as is but the continue isn't too useful and you could combine the cases

PR #201, opened 2022-12-08
[OSE-177] Added GetOperation and ReviewSubmission impl

andresuribe87:
# Overview
This includes the implementation of the ReviewSubmission, and the GetOperation endpoints. 

# Description
Most of the meat and bones of this is in the `pkg/storage/bolt.go` file, where I made changes to operations and submissions are updated in a single transaction. Noteworthy items:
* Key interfaces that were introduced in order to ensure that storage layer was decoupled from the service layer.
* Moved the `pkg/service/presentation/model.go` into it's own package, to avoid import cycles.

# How Has This Been Tested?
See all test files modified. A key aspect is that once the operation is done, the `GetOperation` will return something like below:
```json
{
  "id": "presentations/submissions/1237489",
  "done": true,
  "result": {
    "response": {
      "status":"approved",
      "reason":"fancy operator said it was ok",
      "id": ...
      "definition_id":...
      "definition_map":...
    }
  }
}
```

# Checklist

## References
SIP6


pkg/server/router/operation.go
```diff
@@ -125,21 +138,38 @@ func (pdr OperationRouter) GetOperations(ctx context.Context, w http.ResponseWri
...
+		Result: OperationResult{
+			Error:    op.Result.Error,
+			Response: serviceToRouterType(op.Result.Response),
+		},
+	}
+}
+
+func serviceToRouterType(result interface{}) any {
+	switch result.(type) {
+	case model.Submission:
```
decentralgabe, line 166: what's the point of this method?

also haven't seen the any type - how is it different than `interface{}`?

pkg/server/server_operation_test.go
```diff
@@ -16,6 +17,83 @@ import (
 )
 
 func TestOperationsAPI(t *testing.T) {
+	t.Run("Marks operation as done after reviewing submission", func(t *testing.T) {
+		s, err := storage.NewBoltDB()
```
decentralgabe, line 21: is there teardown code for this?

pkg/server/server_operation_test.go
```diff
@@ -16,6 +17,83 @@ import (
...
+		err = opRouter.GetOperation(newRequestContextWithParams(map[string]string{"id": createdID}), w, req)
+
+		assert.NoError(t, err)
+		var resp router.Operation
+		assert.NoError(t, json.NewDecoder(w.Body).Decode(&resp))
+		assert.True(t, resp.Done)
+		assert.Empty(t, resp.Result.Error)
+		data, err := json.Marshal(submission)
+		assert.NoError(t, err)
+		var responseAsMap interface{}
```
decentralgabe, line 47: nit: if it's a map consider `map[string]interface{}`

pkg/service/operation/model.go
```diff
@@ -1,13 +1,15 @@
...
+	"github.com/TBD54566975/ssi-sdk/util"
 	"go.einride.tech/aip/filtering"
 	"strings"
 )
 
 type Result struct {
-	Error    string      `json:"error,omitempty"`
-	Response interface{} `json:"response,omitempty"`
+	Error    string `json:"error,omitempty"`
+	Response any    `json:"response,omitempty"`
```
decentralgabe, line 12: mucher nicer than interface{}!

pkg/service/operation/service.go
```diff
@@ -43,19 +47,51 @@ func (s Service) GetOperations(request GetOperationsRequest) (*GetOperationsResp
...
+		ID:   op.ID,
+		Done: op.Done,
+		Result: Result{
+			Error: op.Error,
+		},
+	}
+
+	if len(op.Response) > 0 {
+		switch {
+		case strings.Contains(op.ID, "submissions"):
```
decentralgabe, line 69: could this be an operation constant somewhere to avoid collisions in the future?

pkg/service/operation/service.go
```diff
@@ -43,19 +47,51 @@ func (s Service) GetOperations(request GetOperationsRequest) (*GetOperationsResp
...
+			var s prestorage.StoredSubmission
+			if err := json.Unmarshal(op.Response, &s); err != nil {
```
decentralgabe, line 71: nit: consider returning an Operation pointer to avoid this

pkg/service/operation/storage/bolt.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 const (
-	namespace = "operation"
+	Namespace = "operation"
```
decentralgabe, line 13: why does this need to be exported?

pkg/service/presentation/storage/bolt.go
```diff
@@ -20,6 +20,63 @@ type BoltPresentationStorage struct {
 	db *storage.BoltDB
 }
 
+type opUpdater struct {
+	storage.UpdaterWithMap
+}
+
+func (u opUpdater) SetUpdatedResponse(first []byte) {
```
decentralgabe, line 27: why first?

decentralgabe: maybe `response []byte`?

pkg/service/presentation/storage/bolt.go
```diff
@@ -20,6 +20,63 @@ type BoltPresentationStorage struct {
...
+	}
+
+	if op.Done {
+		return errors.New("operation already marked as done")
+	}
+
+	return nil
+}
+
+var _ storage.ResponseSettingUpdater = (*opUpdater)(nil)
```
decentralgabe, line 44: never quite got the point of these can you help me understand?

decentralgabe: do you get a compile err without it?

pkg/service/presentation/storage/bolt.go
```diff
@@ -20,6 +20,63 @@ type BoltPresentationStorage struct {
...
+		storage.NewUpdater(m),
+		opstorage.Namespace,
+		opID,
+		opUpdater{
+			storage.NewUpdater(map[string]any{
+				"done": true,
+			}),
+		})
+	if err != nil {
+		return StoredSubmission{}, opstorage.StoredOperation{}, errors.Wrap(err, "updating value and operation")
```
decentralgabe, line 66: nit: consider returning pointers to avoid needing to instantiate empty objects

pkg/storage/bolt.go
```diff
@@ -141,6 +142,108 @@ func (b *BoltDB) DeleteNamespace(namespace string) error {
...
 }
 
+// UpdaterWithMap is a json map based Updater implementation. The key/values from the map are used to update the
+// unmarshalled JSON representation of the stored data.
+type UpdaterWithMap struct {
+	Values map[string]any
+}
+
+func (u UpdaterWithMap) Validate(v []byte) error {
+	return nil
```
decentralgabe, line 152: hmm, is there any validation that can be done here?

pkg/storage/bolt.go
```diff
@@ -141,6 +142,108 @@ func (b *BoltDB) DeleteNamespace(namespace string) error {
...
+	// SetUpdatedResponse sets the response that the Update method will later use to modify the data.
+	SetUpdatedResponse([]byte)
+}
+
+// UpdateValueAndOperation updates the value stored in (namespace,key) with the new values specified in the map.
+// The updated value is then stored inside the (opNamespace, opKey), and the "done" value is set to true.
+func (b *BoltDB) UpdateValueAndOperation(namespace, key string, updater Updater, opNamespace, opKey string, opUpdater ResponseSettingUpdater) (first, op []byte, err error) {
+	err = b.db.Update(func(tx *bolt.Tx) error {
+		err := updateTxFn(namespace, key, updater, &first)(tx)
+		if err != nil {
```
decentralgabe, line 195: nit: one liner here and below

pkg/server/server_operation_test.go
decentralgabe: should this be TestBoltDB

decentralgabe: nice

pkg/service/operation/service.go
decentralgabe: my stylistic preference but no dealbreaker 😄

pkg/service/operation/storage/bolt.go
decentralgabe: ahh gotcha

pkg/service/presentation/storage/bolt.go
decentralgabe: thanks - makes sense. should probably get this pattern in use in a few other places for safety

pkg/service/presentation/storage/bolt.go
decentralgabe: I think there's a semantic difference between an empty object and a nil object (e.g. empty set vs no set at all)

I agree it makes the caller API a bit more annoying because of the ptr dereferencing issue. I really wish you could get around this in go with an Optional type. There is [a library](https://markphelps.me/posts/option-types-with-go-generate/) but it's fairly un-go-like.

one of my favorite parts of scala was [the option type](https://www.baeldung.com/scala/option-type) which I believe has made its way into java and kotlin too

pkg/storage/bolt.go
decentralgabe: sounds good to me. may be worth adding a comment to capture that for future readers wondering if this can be removed

decentralgabe (approved):
approving pending conflict reviews and linters

pkg/storage/bolt.go
andresuribe87: Added comment.

pkg/service/presentation/storage/bolt.go
andresuribe87: Definitely agree that it's very much not idiomatic go to use option type. Having used them in the past, I view `(*T, error)` in go as exactly the same. 

In Java world (I assume it's similar in Scala), there is a slight different between something like `Optional<int>` and `Optional<Integer>`, since in the latter, the object can be `null`. That's a similar problem as what we're discussing, I think.

PR #203, opened 2022-12-08
[OSE-261] Parent now a required argument when fetching operations

andresuribe87:
# Overview
The `GetOperations` endpoint had an argument called `Parent` which didn't have any effect. This PR implements the expected behavior. 

# Description
The `Parent` argument is now required. Additionally, I change the implementation of the operation service slightly so that operations with different response types end up in different buckets. 

# How Has This Been Tested?
Unit tests

# Checklist

## References
N/A

### DIFF
--- a/pkg/service/operation/storage/bolt.go
+++ b/pkg/service/operation/storage/bolt.go
@@ -1,17 +1,39 @@
 package storage
 
 import (
+	"fmt"
 	"github.com/goccy/go-json"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/tbd54566975/ssi-service/internal/util"
 	"github.com/tbd54566975/ssi-service/pkg/storage"
 	"go.einride.tech/aip/filtering"
+	"strings"
 )
 
 const (
-	namespace = "operation"
+	namespace  = "operation"
+	submission = "submission"
 )
+
+const SubmissionParentResource = "/presentations/submissions"
+
+func namespaceFromID(id string) string {
+	i := strings.LastIndex(id, "/")
+	if i == -1 {
+		return ""
+	}
+	return namespaceFromParent(id[:i])
+}
+
+func namespaceFromParent(parent string) string {
+	switch parent {
+	case SubmissionParentResource:
+		return fmt.Sprintf("%s_%s", namespace, submission)
+	default:
+		return ""
+	}
+}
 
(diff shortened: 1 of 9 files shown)


pkg/service/operation/storage/bolt.go
```diff
@@ -1,18 +1,40 @@
...
+	namespace  = "operation"
+	submission = "submission"
 )
 
+const SubmissionParentResource = "/presentations/submissions"
+
+func namespaceFromID(id string) string {
+	i := strings.LastIndex(id, "/")
+	if i == -1 {
+		return ""
```
nitro-neal, line 24: We might want to panic here, we don't want to somehow write to the namespace ""

pkg/service/operation/storage/bolt.go
```diff
@@ -1,18 +1,40 @@
...
 	"github.com/tbd54566975/ssi-service/internal/util"
 	"github.com/tbd54566975/ssi-service/pkg/storage"
 	"go.einride.tech/aip/filtering"
+	"strings"
 )
 
 const (
-	namespace = "operation"
+	namespace  = "operation"
+	submission = "submission"
```
nitro-neal, line 16: I'm not sure why we need different buckets?
Why can't all operations live in the operations bucket?

If we do need a submission bucket can we name it operations-submission or something so that we know there is no overlap with presentation bucket

nitro-neal (approved):

pkg/service/operation/storage/bolt.go
andresuribe87: I think this is ok because of https://github.com/boltdb/bolt/blob/master/bucket.go#L167

A panic would kill the server, so it's better to avoid it.

pkg/service/operation/storage/bolt.go
andresuribe87: > Why can't all operations live in the operations bucket?

They could! I decided to have separate buckets per operation response type because I expect that's going to be way in which they get queried. I imagine that the UI will have the following 2 screens:
* Show me all pending operations related to submissions;
* Show me all pending operations related to credential applications;

The bucket name actually gets resolved into `operation_submission`. see https://github.com/TBD54566975/ssi-service/pull/203/files#diff-4fbdd974b372fe74a569df1cb0f988c1429d1b4611b84ef6bc9d17af22923e61R29

PR #207, opened 2022-12-09
[chore] Fmt any and goimports

andresuribe87:
# Overview
* Added goimports to the linter. 
* Applied goimports to all files. 
* Renamed `interface{} => any` because it's nicer.


integration/common.go
```diff
@@ -3,20 +3,21 @@ package integration
...
-	"github.com/goccy/go-json"
 	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
 	manifestsdk "github.com/TBD54566975/ssi-sdk/credential/manifest"
 	"github.com/TBD54566975/ssi-sdk/crypto"
+	"github.com/goccy/go-json"
+	cmpact "github.com/goccy/go-json"
```
decentralgabe, line 14: I know you didn't add this but we don't need the alias

decentralgabe (approved):

PR #215, opened 2022-12-12
In memory storage implementation

andresuribe87:
Implemented an in-memory storage conforming to the `ServiceStorage` interface. As part of this, some additional methods were added to the `ServiceStorage`.

Additionally, I changed all tests so they us the in-memory implementation. This makes it simpler to reinstantiate tests.


decentralgabe:
Closing as outdated

Issue #193, opened 2022-12-01
Revocation Concurrency Issue


nitro-neal:
There is a race condition that can occur when you are updating revocations.

vc 1
vc 2
cred status list 1

Thread 1
update vc 1 
update cred status list 1

Thread 2
update vc 2
update cred status list 1
save cred status list 1

Thread 1
save cred status list 1

Thread 2 got overwritten basically and there is no revocation for vc2

Need to make this updateStatus api atomic

nitro-neal:
Fixed here - https://github.com/TBD54566975/ssi-service/commit/c474c091861c40e74c339ec5ef35cba09025c456

PR #217, opened 2022-12-13
Revocation multithread

nitro-neal:
# Overview
There is a race condition that can occur when you are updating revocations.

vc 1
vc 2
cred status list 1

Thread 1
update vc 1
update cred status list 1

Thread 2
update vc 2
update cred status list 1
save cred status list 1

Thread 1
save cred status list 1

Thread 2 got overwritten basically and there is no revocation for vc2

Need to make this updateStatus api atomic

https://github.com/TBD54566975/ssi-service/issues/193

# Description
This change puts a mutex around the updateCredStatus function 

# How Has This Been Tested?
New integration test

# Checklist

- [X ] I have read the CONTRIBUTING document.
- [ X] My code is consistent with the rest of the project 
- [X ] I have tagged the relevant reviewers and/or interested parties
- [X ] I have updated the READMEs and other documentation of affected packages


pkg/server/router/credential.go
```diff
@@ -238,6 +238,8 @@ type UpdateCredentialStatusResponse struct {
 	Revoked bool `json:"revoked"`
 }
 
+var mutex sync.Mutex
```
andresuribe87, line 241: I believe this wouldn't work if there are multiple instances of `ssi-service` running with the same backing DB. 

Have you considered using DB transactions instead to update the credential status? You can see how https://github.com/andresuribe87/ssi-service/blob/8f69172d96017ba9f3bcb1ee263d1b6a412a12c7/pkg/storage/bolt.go#L207 does it for an example.

andresuribe87 (changes requested):

pkg/server/router/credential.go
nitro-neal: that is true, 🤔 

in our upcoming db interface change to have the ability connect to many dbs we will have to account for this in all dbs (some dbs may not support this)

In this case of unpredictable dbs it may be better to do it at the service level with mutex's. 

What do you think?

nitro-neal: maybe have it in both spots to try to account for both scenarios ?

decentralgabe: i think going with a txns model makes sense...whatever solution we go with we'll need that capability across dbs

pkg/server/router/credential.go
```diff
@@ -250,6 +253,9 @@ type UpdateCredentialStatusResponse struct {
 // @Failure      500      {string}  string  "Internal server error"
 // @Router       /v1/credentials/{id}/status [put]
 func (cr CredentialRouter) UpdateCredentialStatus(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
+	// Acquire the mutex lock.
+	mutex.Lock()
```
decentralgabe, line 257: we should keep db concerns at the db layer 😄

PR #218, opened 2022-12-14
[OSE-179] Integration tests for presentation exchange flow

andresuribe87:
# Overview
This follows a similar pattern as all other integration tests. 

# Description
Noteworthy:
* the id of operations was changed from `/presentations/submissions/<uuid>` to `presentations/submissions/<uuid>` (no leading `/`). This was in order to make it easier to pluck out the `id` of the presentation from the URL path.
* Changed the cancel endpoint from `v1/operations/presentations/submissions/{id}/cancel` to `v1/operations/cancel/presentations/submissions/{id}`. The reason was because I found no reasonable way to encode the first one using `httptreemux`. Any suggestions are welcome. 
* Note that using `*id` acts as a wildcard when registering handlers.

### DIFF
--- a/integration/common.go
+++ b/integration/common.go
@@ -11,6 +11,7 @@ import (
 	manifestsdk "github.com/TBD54566975/ssi-sdk/credential/manifest"
 	"github.com/TBD54566975/ssi-sdk/crypto"
 	"github.com/goccy/go-json"
+	"github.com/google/uuid"
 	"github.com/mr-tron/base58"
 	"github.com/oliveagle/jsonpath"
 	"github.com/pkg/errors"
@@ -91,6 +92,22 @@ func CreateVerifiableCredential(issuerDID, schemaID string, revocable bool) (str
 	return output, nil
 }
 
+func CreateSubmissionCredential(issuerDID, subjectDID string) (string, error) {
+	logrus.Println("\n\nCreate a submission credential")
+
+	credentialJSON := getJSONFromFile("submission-credential-input.json")
+
+	credentialJSON = strings.ReplaceAll(credentialJSON, "<CREDISSUERID>", issuerDID)
+	credentialJSON = strings.ReplaceAll(credentialJSON, "<CREDSUBJECTID>", subjectDID)
+
+	output, err := put(endpoint+version+"credentials", credentialJSON)
+	if err != nil {
+		return "", errors.Wrapf(err, "credentials endpoint with output: %s", output)
+	}
+
+	return output, nil
+}
+
 func CreateCredentialManifest(issuerDID, schemaID string) (string, error) {
 	logrus.Println("\n\nCreate our Credential Manifest:")
 	manifestJSON := getJSONFromFile("manifest-input.json")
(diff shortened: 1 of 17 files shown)


integration/common.go
```diff
@@ -136,6 +153,74 @@ func CreateCredentialApplicationJWT(presentationDefinitionID, credentialJWT, man
...
+	}
+
+	return output, nil
+}
+
+func CreateSubmission(definitionID, holderDID, holderPrivateKey, credentialJWT string) (string, error) {
+	logrus.Println("\n\nCreate our Submission:")
+	submissionJSON := getJSONFromFile("presentation-submission-input.json")
+
+	submissionJSON = strings.ReplaceAll(submissionJSON, "<DEFINITIONID>", definitionID)
```
decentralgabe, line 182: would really like us to switch to go templating -- https://pkg.go.dev/text/template

integration/testdata/review-submission-input.json
```diff
@@ -0,0 +1,4 @@
+{
+  "approved":true,
```
decentralgabe, line 2: nit: json formatting

pkg/server/server.go
```diff
@@ -212,9 +212,9 @@ func (s *SSIServer) OperationAPI(service svcframework.Service) (err error) {
 
 	handlerPath := V1Prefix + OperationPrefix
 
-	s.Handle(http.MethodGet, path.Join(handlerPath, "/:id"), operationRouter.GetOperation)
 	s.Handle(http.MethodGet, handlerPath, operationRouter.GetOperations)
-	s.Handle(http.MethodPut, path.Join(handlerPath, "/:id/cancel"), operationRouter.CancelOperation)
+	s.Handle(http.MethodPut, path.Join(handlerPath, "/cancel/*id"), operationRouter.CancelOperation)
```
decentralgabe, line 216: I haven't seen this syntax, why *id instead of :id?

andresuribe87: From https://github.com/dimfeld/httptreemux#routing-rules

> Path elements starting with : indicate a wildcard in the path. A wildcard will only match on a single path segment. That is, the pattern /post/:postid will match on /post/1 or /post/1/, but not /post/1/2.
> 
> A path element starting with * is a catch-all, whose value will be a string containing all text in the URL matched by the wildcards. For example, with a pattern of /images/*path and a requested URL images/abc/def, path would contain abc/def. A catch-all path will not match an empty string, so in this example a separate route would need to be installed if you also want to match /images/.

In this case, I use it so that the operation id matches `presentations/submissions/{submission_id}` for the URL path 
 `v1/operations/cancel/presentations/submissions/{id}`

integration/testdata/review-submission-input.json
andresuribe87: Done.

integration/common.go
andresuribe87: Great idea. Will take that in a separate PR.

decentralgabe: ty

pkg/server/server.go
decentralgabe: gotcha! might want to add a comment to show how this is used here

decentralgabe (approved):

PR #221, opened 2022-12-15
Make stuff fast

andresuribe87:
This enables caching for tests, and disables -race from Actually enable caching for tests.

I'm not convinced we should do the latter part, this was just an example.


decentralgabe (approved):
🚤

nitro-neal:
👍

PR #222, opened 2022-12-16
Adding Redis DB

nitro-neal:
# Overview
Adding Redis DB

# Description
This change adds redis db with a mutex on writes as a configuration value in our toml config file

# How Has This Been Tested?
image

# Checklist

- [ ✅ ] I have read the CONTRIBUTING document.
- [ ✅ ] My code is consistent with the rest of the project 
- [ ✅ ] I have tagged the relevant reviewers and/or interested parties
- [ ✅ ] I have updated the READMEs and other documentation of affected packages

## References
https://forums.tbd.website/t/ssi-service-database-considerations/216/3


nitro-neal:
We have decided to use **Redis** (go-redis -[https://redis.uptrace.dev/guide/](https://redis.uptrace.dev/guide/) ) as our first database implementation. It implements all of our requirements and is robust enough to scale and cover future requirements as well

- Super fast low latency
- Redis allows storing key and value pairs as large as 512 MB.
- Transactional Support and ACID compliant
- Huge ecosystem with many plugins
- High availability and scalability
- Open source
- Enterprise Support and available on most cloud platforms

nitro-neal:
If we are onboard with the way this is implemented the next steps would be:

1. Make many acceptance tests making sure we get a 1:1 match from our redis db to our bolt db to our future dbs we would implement.

2. Remove These unused methods from the interface:
	URI() string
	IsOpen() bool
	
3. Refactor / remove these from the interface
	Update(namespace string, key string, values map[string]any) ([]byte, error)
	UpdateValueAndOperation(namespace, key string, updater Updater, opNamespace, opKey string, opUpdater ResponseSettingUpdater) (first, op []byte, err error)
	
as they leak implementation details

pkg/storage/redis.go
```diff
@@ -0,0 +1,216 @@
...
+const RedisScanBatchSize = 1000
+const RedisMutex = "redis-mutex"
+
+func init() {
+	_ = RegisterStorage(&RedisDB{})
+}
+
+type RedisDB struct {
+	db    *goredislib.Client
+	ctx   context.Context
```
andresuribe87, line 25: This should likely be the first parameter of all the interface methods, so we can propagate deadlines as well as cancel requests.

pkg/storage/redis.go
```diff
@@ -0,0 +1,216 @@
...
+	mutex *redsync.Mutex
+}
+
+func (b *RedisDB) Init(i interface{}) error {
+	options := i.(map[string]interface{})
+
+	client := goredislib.NewClient(&goredislib.Options{
+		Addr:     options["address"].(string),
+		Password: options["password"].(string),
```
andresuribe87, line 34: Passwords will likely need to be injected via environment variables.

pkg/storage/redis.go
```diff
@@ -0,0 +1,216 @@
...
+		return errors.Wrap(err, "cannot obtain mutex lock")
+	}
+	defer func() {
+		ok, unlockErr := b.mutex.Unlock()
+		if !ok || unlockErr != nil {
+			logrus.Error(unlockErr)
+		}
+	}()
+
+	return b.db.Set(b.ctx, nameSpaceKey, value, 0).Err()
```
andresuribe87, line 89: A comment on why the 0 would be nice :)

pkg/storage/redis.go
```diff
@@ -0,0 +1,216 @@
...
+
+func readAllKeys(match string, b *RedisDB) ([]string, error) {
+	var cursor uint64
+
+	allKeys := make([]string, 0)
+
+	for {
+		keys, nextCursor, err := b.db.Scan(b.ctx, cursor, match+"*", RedisScanBatchSize).Result()
+		if err != nil {
+			return nil, errors.Wrap(err, "redis scan error")
```
andresuribe87, line 154:
```suggestion
			return nil, errors.Wrap(err, "scan error")
```

pkg/storage/redis.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	return allKeys, nil
+}
+
+func (b *RedisDB) Delete(namespace, key string) error {
+	nameSpaceKey := getRedisKey(namespace, key)
+
+	if err := b.mutex.Lock(); err != nil {
+		return errors.Wrap(err, "cannot obtain mutex lock")
```
andresuribe87, line 173:
```suggestion
		return errors.Wrap(err, "locking")
```

pkg/storage/redis.go
```diff
@@ -0,0 +1,216 @@
...
+func readAll(keys []string, b *RedisDB) (map[string][]byte, error) {
+	result := make(map[string][]byte)
+
+	if len(keys) == 0 {
+		return nil, nil
+	}
+
+	values, err := b.db.MGet(b.ctx, keys...).Result()
+	if err != nil {
+		return nil, errors.Wrap(err, "mget error, read all error")
```
andresuribe87, line 126:
```suggestion
		return nil, errors.Wrap(err, "getting multiple keys")
```

pkg/storage/redis.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	if len(keys) != len(values) {
+		panic("key length does not equal value length")
```
andresuribe87, line 130: Perhaps it's better to return an error? https://github.com/uber-go/guide/blob/master/style.md#dont-panic

pkg/storage/redis.go
```diff
@@ -0,0 +1,216 @@
...
+		}
+	}()
+
+	return b.db.Del(b.ctx, nameSpaceKey).Err()
+}
+
+func (b *RedisDB) DeleteNamespace(namespace string) error {
+	keys, err := readAllKeys(namespace, b)
+	if err != nil {
+		return errors.Wrap(err, "read all keys error")
```
andresuribe87, line 188:
```suggestion
		return errors.Wrap(err, "read all keys")
```

pkg/storage/redis.go
```diff
@@ -0,0 +1,216 @@
...
+		if !ok || unlockErr != nil {
+			logrus.Error(unlockErr)
+		}
+	}()
+
+	return b.db.Del(b.ctx, nameSpaceKey).Err()
+}
+
+func (b *RedisDB) DeleteNamespace(namespace string) error {
+	keys, err := readAllKeys(namespace, b)
```
andresuribe87, line 186: Maybe add a TODO here that this might be dangerous as it might run out of memory?

pkg/storage/redis.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+// TODO: Implement
+func (b *RedisDB) Update(namespace string, key string, values map[string]any) ([]byte, error) {
+	panic("implement me")
+}
+
+// TODO Implement
+func (b *RedisDB) UpdateValueAndOperation(namespace, key string, updater Updater, opNamespace, opKey string, opUpdater ResponseSettingUpdater) (first, op []byte, err error) {
+	panic("implement me")
```
andresuribe87, line 211: This method is actually used for operations, so we shouldn't remove it.

pkg/storage/redis.go
```diff
@@ -0,0 +1,216 @@
...
+
+	goredislib "github.com/go-redis/redis/v8"
+	"github.com/go-redsync/redsync/v4"
+	"github.com/go-redsync/redsync/v4/redis/goredis/v8"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+const PONG = "PONG"
```
decentralgabe, line 15: nit: const block

decentralgabe (approved):
this is awesome.

adding @michaelneale for review

andresuribe87 (changes requested):

pkg/storage/redis.go
```diff
@@ -0,0 +1,216 @@
...
+const RedisScanBatchSize = 1000
+const RedisMutex = "redis-mutex"
+
+func init() {
+	_ = RegisterStorage(&RedisDB{})
```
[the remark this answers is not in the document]
nitro-neal: yea I'll add panic here

pkg/storage/redis.go
nitro-neal: Looking at the comment:
// UpdateValueAndOperation updates the value stored in (namespace,key) with the new values specified in the map.
// The updated value is then stored inside the (opNamespace, opKey), and the "done" value is set to true.

It seems the database layer needs to know things about the service

Shouldn't we reimplement this to maybe be like an array of updates to key value pairs and it update both or none at all?

Something like this:
func (b *RedisDB) WriteMany(namespaces []string, key []string, value [][]byte) error

pkg/storage/redis.go
nitro-neal: Yea if this is a common golang thing, sounds good. 
So no matter what db we use mongo / bolt / couch they can all use the same context object ?

nitro-neal: image
You are saying to add context as the first param to all these functions?

And then many places throughout the codebase they will have to access this context, is there a good way to do this?

andresuribe87: Every incoming request should create a context. That should be passed along throughout the stack for any operations that do IO. Cancelations  are done using it.

pkg/storage/redis.go
andresuribe87: Ahh good point, that documentation should be updated. That logic you described is actually encoded in the updater object, so it's specified in the application layer.

andresuribe87: Note that the Update method is unused, but the UpdateValueAndOperation is.

michaelneale:
very cool - is redis likely to be durable enough for the typical user of SSI? durability as in sufficiently long term storage? 

If this is helping flesh out a storage api, should we include an additional very different implementation such as s3 that is more on the durability scale and less on the low latency cache?

pkg/storage/redis.go
nitro-neal: Added issue for this - https://github.com/TBD54566975/ssi-service/issues/228

pkg/storage/redis.go
nitro-neal: I think this can be removed and we can rework the code to use the newly implemented WriteMany command, what do you think?

func (b *RedisDB) WriteMany(namespaces, keys []string, values [][]byte) error

andresuribe87: I think that may work, but I'm not sure exactly how. Perhaps you can share what you have in mind? 

I'll note that the existing `UpdateValueAndOperation` takes the value the newly updated value from the first key, and writes that value into the second key. In order to support something like that, we'll need to do a read first and then call WriteMany. That would need to be done within the transaction.

nitro-neal: I have implemented this function .

nitro-neal:
image

pkg/storage/redis.go
```diff
@@ -0,0 +1,366 @@
...
+func init() {
+	err := RegisterStorage(&RedisDB{})
+	if err != nil {
+		panic(err)
+	}
+}
+
+type RedisDB struct {
+	db    *goredislib.Client
+	ctx   context.Context
```
andresuribe87, line 30: Having as an instance variable could potentially be dangerous. E.g. if there are changes to it. I recommend removing. I recommend removing for now, and creating a new context on every request.

We'll fix this later on to do it the right way.

pkg/storage/redis.go
```diff
@@ -0,0 +1,366 @@
...
+	b.db = client
+	b.mutex = rs.NewMutex(RedisMutex)
+	b.ctx = context.Background()
+
+	if options["flush"] != nil && options["flush"].(bool) {
+		err := b.db.FlushAll(b.ctx).Err()
+		if err != nil {
+			return err
+		}
+	}
```
andresuribe87, line 54: This seems a bit dangerous, no? If this is for testing, perhaps there is a way of preventing it from happening in production? I can see build tags being used for such a case.

pkg/storage/redis.go
```diff
@@ -0,0 +1,366 @@
...
+
+	if err := b.mutex.Lock(); err != nil {
+		return errors.Wrap(err, "cannot obtain mutex lock")
+	}
+	defer func() {
+		ok, unlockErr := b.mutex.Unlock()
+		if !ok || unlockErr != nil {
+			logrus.Error(unlockErr)
+		}
+	}()
```
andresuribe87, line 92: The `SET` command is an atomic operation in Redis, so locking shouldn't be required for a write operation.
```suggestion
```

pkg/storage/redis.go
```diff
@@ -0,0 +1,366 @@
...
+
+	// Zero expiration means the key has no expiration time.
+	return b.db.Set(b.ctx, nameSpaceKey, value, 0).Err()
+}
+
+func (b *RedisDB) WriteMany(namespaces, keys []string, values [][]byte) error {
```
andresuribe87, line 98: Is `WriteMany` used?

Separately, did you consider using `MSET` instead of a pipeline? The benefit is that you wouldn't need to lock, and that it's significantly faster.

pkg/storage/redis.go
```diff
@@ -0,0 +1,366 @@
...
+
+	return res, err
+}
+
+func (b *RedisDB) ReadPrefix(namespace, prefix string) (map[string][]byte, error) {
+	namespacePrefix := getRedisKey(namespace, prefix)
+
+	keys, err := readAllKeys(namespacePrefix, b)
+	if err != nil {
+		return nil, errors.Wrap(err, "read all keys error")
```
andresuribe87, line 150:
```suggestion
		return nil, errors.Wrap(err, "read all keys")
```

pkg/storage/redis.go
```diff
@@ -0,0 +1,366 @@
...
+	}
+
+	return readAll(keys, b)
+}
+
+func (b *RedisDB) ReadAll(namespace string) (map[string][]byte, error) {
+	keys, err := readAllKeys(namespace, b)
+	if err != nil {
+		return nil, errors.Wrap(err, "read all keys error")
```
andresuribe87, line 159:
```suggestion
		return nil, errors.Wrap(err, "read all keys")
```

pkg/storage/redis.go
```diff
@@ -0,0 +1,366 @@
...
+	}
+
+	return readAll(keys, b)
+}
+
+// TODO: This potentially could dangerous as it might run out of memory as we populate result
+func readAll(keys []string, b *RedisDB) (map[string][]byte, error) {
+	result := make(map[string][]byte)
```
andresuribe87, line 167:
```go
result := make(map[string][]byte, len(keys))
```

pkg/storage/redis.go
```diff
@@ -0,0 +1,366 @@
...
+	if len(keys) != len(values) {
+		return nil, errors.New("key length does not match value length")
+	}
+
+	// result needs to take the namespace out of the key
+	namespaceDashIndex := strings.Index(keys[0], "-")
+	for i, val := range values {
+		byteValue := []byte(fmt.Sprintf("%v", val))
+		key := keys[i]
+		key = key[namespaceDashIndex+1:]
```
andresuribe87, line 187:
```suggestion
		key := keys[i][namespaceDashIndex+1:]
```

pkg/storage/redis.go
```diff
@@ -0,0 +1,366 @@
...
+	for i, key := range keys {
+		keyWithoutNamespace := key[namespaceDashIndex+1:]
+		keys[i] = keyWithoutNamespace
+	}
+
+	return keys, nil
+}
+
+// TODO: This potentially could dangerous as it might run out of memory as we populate allKeys
+func readAllKeys(namespace string, b *RedisDB) ([]string, error) {
```
andresuribe87, line 214: Seems like you're removing the namespace out of the keys every time. Would it make sense to do it inside this function?

pkg/storage/redis.go
```diff
@@ -0,0 +1,366 @@
...
+	var cursor uint64
+
+	allKeys := make([]string, 0)
```
andresuribe87, line 217:
```suggestion
	var allKeys []string
```

pkg/storage/redis.go
```diff
@@ -0,0 +1,366 @@
...
+		}
+	}()
+
+	if !namespaceExists(namespace, b) {
+		return fmt.Errorf("namespace<%s> does not exist", namespace)
+	}
+
+	res, err := b.db.GetDel(b.ctx, nameSpaceKey).Result()
+	if res == "" {
+		return errors.Wrap(err, fmt.Sprintf("key<%s> and namespace<%s> does not exist", key, namespace))
```
andresuribe87, line 256:
```suggestion
		return errors.Wrapf(err, "key<%s> and namespace<%s> does not exist", key, namespace)
```

pkg/storage/redis.go
```diff
@@ -0,0 +1,366 @@
...
+
+	if err := b.mutex.Lock(); err != nil {
+		return errors.Wrap(err, "locking")
+	}
+	defer func() {
+		ok, unlockErr := b.mutex.Unlock()
+		if !ok || unlockErr != nil {
+			logrus.Error(unlockErr)
+		}
+	}()
```
andresuribe87, line 248: The command `GETDEL` is atomic, so I believe this isn't needed.

pkg/storage/redis.go
```diff
@@ -0,0 +1,366 @@
...
+
+	if err := b.mutex.Lock(); err != nil {
+		return errors.Wrap(err, "cannot obtain mutex lock")
+	}
+	defer func() {
+		ok, unlockErr := b.mutex.Unlock()
+		if !ok || unlockErr != nil {
+			logrus.Error(unlockErr)
+		}
+	}()
```
andresuribe87, line 277: I think this isn't needed either, since `DEL` is atomic.

pkg/storage/redis.go
```diff
@@ -0,0 +1,366 @@
...
+
+	if len(keys) == 0 {
+		return fmt.Errorf("could not delete namespace<%s>, namespace does not exist", namespace)
```
andresuribe87, line 280:
```suggestion
		return errors.Errorf("could not delete namespace<%s>, namespace does not exist", namespace)
```

pkg/storage/redis.go
```diff
@@ -0,0 +1,366 @@
...
+func (b *RedisDB) UpdateValueAndOperation(namespace, key string, updater Updater, opNamespace, opKey string, opUpdater ResponseSettingUpdater) (first, op []byte, err error) {
+	if err := b.mutex.Lock(); err != nil {
+		return nil, nil, errors.Wrap(err, "cannot obtain mutex lock")
+	}
+	defer func() {
+		ok, unlockErr := b.mutex.Unlock()
+		if !ok || unlockErr != nil {
+			logrus.Error(unlockErr)
+		}
+	}()
```
andresuribe87, line 300: Since you're executing in a pipeline, is this mutex necessary?

pkg/storage/redis.go
```diff
@@ -0,0 +1,366 @@
...
+}
+
+func txWithUpdater(namespace, key string, updater Updater, b *RedisDB) ([]byte, error) {
+	nameSpaceKey := getRedisKey(namespace, key)
+	v, err := b.db.Get(b.ctx, nameSpaceKey).Bytes()
+	if err != nil {
+		return nil, errors.Wrapf(err, "get error with namespace: %s key: %s", namespace, key)
+	}
+	if v == nil {
+		return nil, fmt.Errorf("key not found %s", key)
```
andresuribe87, line 336:
```suggestion
		return nil, errors.Errorf("key not found %s", key)
```

pkg/storage/redis.go
```diff
@@ -0,0 +1,366 @@
...
+	defer func() {
+		ok, unlockErr := b.mutex.Unlock()
+		if !ok || unlockErr != nil {
+			logrus.Error(unlockErr)
+		}
+	}()
+
+	if !namespaceExists(namespace, b) {
+		return fmt.Errorf("namespace<%s> does not exist", namespace)
+	}
```
andresuribe87, line 252: Curious if this check could be removed?

andresuribe87 (commented):
This is looking great! Some minor comments to hopefully remove the need to do distributed locking. 

Separately, consider also testing all the `server` package files. I did a gross prototype of that in this PR https://github.com/TBD54566975/ssi-service/pull/224/files#r1055833324

pkg/storage/redis.go
nitro-neal: If I remove this check it will not be 1:1 sync with the boltdb implementation :)
There are a few places like this where we do a few 'not needed' things to match up with the bolt impl

config/compose.toml
```diff
@@ -21,9 +21,20 @@ log_level = "info"
...
+
+# Redis Configuration
+# [services]
+# service_endpoint = "http://localhost:8080"
+# storage = "redis"
+
+# [services.storage_option]
+# address = "redis:6379"
+# password = ""
+# flush = true
```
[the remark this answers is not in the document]
nitro-neal: yea maybe I can make a compose.toml.example ? to leave this in to show how redis will be hooked up?

pkg/storage/redis.go
nitro-neal: I'll take this flush option out and have it flush in our docker compose

pkg/storage/redis.go
nitro-neal: Ok well in that case we almost don't need a mutex. Does the UpdateValueAndOperation even still need it?

pkg/storage/redis.go
nitro-neal: Nice:
MSET is atomic, so all given keys are set at once. It is not possible for clients to see that some of the keys were updated while others are unchanged.

ok I can just use that.

Redis Rocks

build/docker-compose.yml
```diff
@@ -10,8 +11,11 @@ services:
       - JAEGER_HTTP_URL=http://jaeger:14268/api/traces
     depends_on:
       - jaeger
+      - redis
     volumes:
       - ../config/compose.toml:/app/config/config.toml
+    links:
+      - redis
```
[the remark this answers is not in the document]
nitro-neal: Created new issue - https://github.com/decentralized-identity/credential-manifest/issues/144

config/compose.toml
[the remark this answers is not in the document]
andresuribe87: Yeah that works. An example in the readme could also work!

andresuribe87 (approved):

config/compose.toml
```diff
@@ -21,9 +21,10 @@ log_level = "info"
 
 enable_schema_caching = true
 
+# Bolt Configuration
 [services]
-storage = "bolt"
 service_endpoint = "http://localhost:8080"
```
decentralgabe, line 26: let's think how this plays in with docker compose. i believe we should consider reading in config through env vars as an override for config set here, where docker compose would set this as an env var

pkg/storage/db_test.go
```diff
@@ -0,0 +1,374 @@
...
+func getDBImplementations(t *testing.T) []ServiceStorage {
+	boltDB := setupBoltDB(t)
+	redisDB := setupRedisDB(t)
+
+	dbImpls := make([]ServiceStorage, 0)
+	dbImpls = append(dbImpls, boltDB, redisDB)
+	return dbImpls
+}
+
+func setupBoltDB(t *testing.T) *BoltDB {
```
decentralgabe, line 23: what is the value in testing multiple DB implementations at the same time?

do you foresee us needing to run multiple DBs in a real environment?

nitro-neal: This guarantees that the same input, error messages, and output happen with any implementation of any db we decide to implement.

config/compose.toml
nitro-neal: created issue for this - https://github.com/TBD54566975/ssi-service/issues/233

PR #223, opened 2022-12-16
[Chore] Move from string replacement to templating

andresuribe87:
Nice to have for integration tests.


decentralgabe (approved):
Thanks for doing this!

PR #224, opened 2022-12-16
In Memory Storage Implementation

andresuribe87:
# Overview
This is a simple implementation of a storage backend that works in memory.

# Description
This was done mostly as a proof of concept to start hardening the interfaces and explore how we should set up testing with different database providers. 

# How Has This Been Tested?
All the tests in the packages `pkg/server` and `pkg/storage` are run against the new implementation, as well as the Bolt implementation.
Some potential followups: do DB cleanup depending on the implementation.
 
# Checklist

## References
N/A

### DIFF
--- a/pkg/server/server_test.go
+++ b/pkg/server/server_test.go
@@ -34,9 +34,17 @@ import (
 	"github.com/tbd54566975/ssi-service/pkg/storage"
 )
 
+var storageProvider storage.Type
+
 func TestMain(t *testing.M) {
 	testutil.EnableSchemaCaching()
-	os.Exit(t.Run())
+	providersToTest := []storage.Type{storage.Bolt, storage.InMemory}
+	for i, p := range providersToTest {
+		storageProvider = p
+		if c := t.Run(); c != 0 || i == (len(providersToTest)-1) {
+			os.Exit(c)
+		}
+	}
 }
 
 func TestHealthCheckAPI(t *testing.T) {
--- a/pkg/storage/memory.go
+++ b/pkg/storage/memory.go
@@ -1,0 +1,185 @@
+package storage
+
+import (
+	"strings"
+	"sync"
+
+	"github.com/pkg/errors"
+)
+
+const (
+	InMemory Type = "in-memory"
+)
+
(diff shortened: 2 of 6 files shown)


pkg/storage/memory.go
```diff
@@ -0,0 +1,185 @@
...
+func init() {
+	_ = RegisterStorage(&MemoryDB{})
+}
+
+// MemoryDB is an in memory implementation of ServiceStorage that is safe for concurrent use.
+type MemoryDB struct {
+	maps sync.Map
+}
+
+func (f *MemoryDB) Init(interface{}) error {
```
decentralgabe, line 23: nit: `m` instead of `f`

decentralgabe (approved):
looks fine - but what's the value of this over bolt?

andresuribe87:
> looks fine - but what's the value of this over bolt?

This may be useful for testing purposes (it's cleaner than cleaning up the DB). Beyond that, no more value. I'm going to think whether it makes sense to keep.

pkg/server/server_test.go
```diff
@@ -34,9 +34,17 @@ import (
 	"github.com/tbd54566975/ssi-service/pkg/storage"
 )
 
+var storageProvider storage.Type
+
 func TestMain(t *testing.M) {
 	testutil.EnableSchemaCaching()
-	os.Exit(t.Run())
+	providersToTest := []storage.Type{storage.Bolt, storage.InMemory}
```
andresuribe87, line 41: A bit gross, but very effective way of making sure all the tests in the package pass with the new implementation.

decentralgabe:
was thinking about this a bit more and I think we should not add memory and just stick to bolt. I had intended bolt to be the local implantation. it should be light weight and simple for all local environments. if it's not and we need this, let's re-consider bolt.

andresuribe87:
> was thinking about this a bit more and I think we should not add memory and just stick to bolt. I had intended bolt to be the local implantation. it should be light weight and simple for all local environments. if it's not and we need this, let's re-consider bolt.

Yes, I agree with this. Closing out.

PR #225, opened 2022-12-17
well organized service directories.

suutaku:
# Overview
* use environment variable `SSI_HOME` to set home dir of service.
if `SSI_HOME` not set, `./` will be the home path. **note: it called `os.Chdir`, so service executed every thing inside this home path.**
* create `config`,`logs` and `data` as subdirectory at home dir. 
the `config` for configure files, `logs` for log files, `data ` for databases.
*  if `config.toml` not exist, create it with default values and comments.

# Description
A well organized service directory tree.

# How Has This Been Tested?
_Describe the tests that you ran to verify your changes. Provide instructions for verification._

- [x] Configure Marshal and Unmarshal
- [x] Create directories automaticlly
- [x]  Auto created `config.toml` is same with `pkg/config/config.toml` (added some comments)

# Checklist

## References
_Please list relevant documentation (e.g. tech specs, articles, follow up or related work) relevant to this change, and note if the documentation has been updated._

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -197,7 +197,7 @@ func newTracerProvider(cfg config.SSIServiceConfig) (*sdktrace.TracerProvider, e
 
 func createLogFile(location string) (string, error) {
 	if _, err := os.Stat(location); os.IsNotExist(err) {
-		if err = os.MkdirAll(location, 0766); err != nil {
+		if err = os.MkdirAll(location, os.ModePerm); err != nil {
 			return "", err
 		}
 	}
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -72,7 +72,11 @@ func run() error {
 	}
 	// set log config from config file
 	if cfg.Server.LogLocation != "" {
-		file, err := os.OpenFile(createLogFile(cfg.Server.LogLocation), os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0666)
+		logFilePath, err := createLogFile(cfg.Server.LogLocation)
+		if err != nil {
+			return err
+		}
+		file, err := os.OpenFile(logFilePath, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0666)
 		if err == nil {
 			logrus.SetOutput(file)
 		} else {
@@ -191,13 +195,11 @@ func newTracerProvider(cfg config.SSIServiceConfig) (*sdktrace.TracerProvider, e
 	return tp, nil
 }
 
-func createLogFile(location string) string {
-	_, err := os.Stat(location)
-	if os.IsNotExist(err) {
-		err = os.MkdirAll(location, 0766)
-		if err != nil {
-			panic(err)
+func createLogFile(location string) (string, error) {
+	if _, err := os.Stat(location); os.IsNotExist(err) {
+		if err = os.MkdirAll(location, 0766); err != nil {
+			return "", err
 		}
 	}
-	return location + "/" + config.ServiceName + "-" + time.Now().Format(time.RFC3339) + ".log"
+	return location + "/" + config.ServiceName + "-" + time.Now().Format(time.RFC3339) + ".log", nil
 }
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -56,7 +56,7 @@ func main() {
 
 // startup and shutdown logic
 func run() error {
-	cfg, err := config.LoadConfig(config.DefaultConfigPath)
+	cfg, err := config.LoadConfig(os.Getenv("SSI_HOME"))
 	if err != nil {
 		logrus.Fatalf("could not instantiate config: %s", err.Error())
 	}
@@ -76,7 +76,7 @@ func run() error {
 		if err == nil {
 			logrus.SetOutput(file)
 		} else {
-			logrus.Info("Failed to log to file, using default stdout")
+			logrus.Info("Failed to log to file, using default stdout", err)
 		}
 		defer file.Close()
 	}
@@ -95,9 +95,9 @@ func run() error {
 		}
 	}
 
-	expvar.NewString("build").Set(cfg.Version.SVN)
+	expvar.NewString("build").Set(cfg.Svn)
 
-	logrus.Infof("main: Started : Service initializing : version %q", cfg.Version.SVN)
+	logrus.Infof("main: Started : Service initializing : version %q", cfg.Svn)
 	defer logrus.Info("main: Completed")
 
 	out, err := conf.String(cfg)
@@ -185,12 +185,19 @@ func newTracerProvider(cfg config.SSIServiceConfig) (*sdktrace.TracerProvider, e
 		sdktrace.WithResource(resource.NewWithAttributes(
 			semconv.SchemaURL,
 			semconv.ServiceNameKey.String(config.ServiceName),
-			semconv.ServiceVersionKey.String(cfg.Version.SVN),
+			semconv.ServiceVersionKey.String(cfg.Svn),
 		)),
 	)
 	return tp, nil
 }
 
 func createLogFile(location string) string {
+	_, err := os.Stat(location)
+	if os.IsNotExist(err) {
+		err = os.MkdirAll(location, 0766)
+		if err != nil {
+			panic(err)
+		}
+	}
 	return location + "/" + config.ServiceName + "-" + time.Now().Format(time.RFC3339) + ".log"
 }
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -80,7 +80,7 @@ func run() error {
 		if err == nil {
 			logrus.SetOutput(file)
 		} else {
-			logrus.Info("Failed to log to file, using default stdout", err)
+			logrus.WithError(err).Error("Failed to log to file, using default stdout")
 		}
 		defer file.Close()
 	}
--- a/config/config.go
+++ b/config/config.go
@@ -11,7 +11,7 @@ import (
 
 const (
 	defaultConfigPath = "config"
-	ConfigFileName    = "config.toml"
+	configFileName    = "config.toml"
 	ServiceName       = "ssi-service"
 	ConfigExtension   = ".toml"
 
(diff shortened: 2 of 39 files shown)


suutaku:
BTW, why log use JSONFormat? it's hard to read...

cmd/main.go
```diff
@@ -56,7 +56,7 @@ func main() {
 
 // startup and shutdown logic
 func run() error {
-	cfg, err := config.LoadConfig(config.DefaultConfigPath)
+	cfg, err := config.LoadConfig(os.Getenv("SSI_HOME"))
```
decentralgabe, line 59: why is this preferable over a configuration-driven path?

cmd/main.go
```diff
@@ -76,7 +76,7 @@ func run() error {
 		if err == nil {
 			logrus.SetOutput(file)
 		} else {
-			logrus.Info("Failed to log to file, using default stdout")
+			logrus.Info("Failed to log to file, using default stdout", err)
```
decentralgabe, line 79:
```suggestion
			logrus.WithError(err).Error("Failed to log to file, using default stdout")
```

cmd/main.go
```diff
@@ -185,12 +185,19 @@ func newTracerProvider(cfg config.SSIServiceConfig) (*sdktrace.TracerProvider, e
...
 	return tp, nil
 }
 
 func createLogFile(location string) string {
+	_, err := os.Stat(location)
+	if os.IsNotExist(err) {
+		err = os.MkdirAll(location, 0766)
+		if err != nil {
+			panic(err)
+		}
```
decentralgabe, line 200: two things - can combine the err check into one line and let's not panic here and instead return an err

config/config.go
```diff
@@ -1,58 +1,60 @@
...
-	WriteTimeout        time.Duration `toml:"write_timeout" conf:"default:5s"`
-	ShutdownTimeout     time.Duration `toml:"shutdown_timeout" conf:"default:5s"`
-	LogLocation         string        `toml:"log_location" conf:"default:log"`
-	LogLevel            string        `toml:"log_level" conf:"default:debug"`
-	EnableSchemaCaching bool          `toml:"enable_schema_caching" conf:"default:true"`
+	APIHost             string        `toml:"api_host"`
+	DebugHost           string        `toml:"debug_host"`
+	JagerHost           string        `toml:"jager_host"`
+	JagerEnabled        bool          `toml:"jager_enabled"`
+	ReadTimeout         time.Duration `toml:"read_timeout" comment:"5 seconds, time is in nanoseconds"`
```
decentralgabe, line 39: why remove the defaults, I believe they're useful!

cmd/main.go
suutaku: Because in the future there will be more ways to use. now we deploy as a docker image, pull source, build and launch configure.toml from source code. we also can create release binary, event a Linux package. considering the situation, a well organized path is necessary like other application (service)  do(nginx, ssh,etc..).

config/config.go
suutaku: I never removed `default` values, just removed the `default` toml tags.  I have search to find a toml Marshaller which supported comments marshaling and it is `github.com/pelletier/go-toml/v2`.  as it's [README.md](https://github.com/pelletier/go-toml#support-for-default-struct-tag-has-been-dropped) described,**The recommended replacement is pre-filling the struct before unmarshaling.** ,so i modify code as recommended.

cmd/main.go
suutaku: ok. i will fix it.

cmd/main.go
suutaku: whit environment var, we can switch home path easily like command line option way which `github.com/ardanlabs/conf` provided.
but `github.com/ardanlabs/conf`'s marshaller can not marshaling toml comments. considering it's important to create a new, default configure file with comments, i use `github.com/suutaku/nice-config` instead.

cmd/main.go
```diff
@@ -185,12 +189,17 @@ func newTracerProvider(cfg config.SSIServiceConfig) (*sdktrace.TracerProvider, e
...
 		)),
 	)
 	return tp, nil
 }
 
-func createLogFile(location string) string {
-	return location + "/" + config.ServiceName + "-" + time.Now().Format(time.RFC3339) + ".log"
+func createLogFile(location string) (string, error) {
+	if _, err := os.Stat(location); os.IsNotExist(err) {
+		if err = os.MkdirAll(location, 0766); err != nil {
```
andresuribe87, line 200: Perhaps the `os` package consts would make this more readable, instead of 0766.

andresuribe87 (commented):
IMO, we should be moving to a more mature configuration package. Something like [viper](https://github.com/spf13/viper). I've used it in the past to deal to good success. It has facilities to tie in environment variables with config files, and also gives us a path forward towards doing on-the-fly swapping.

suutaku:
> IMO, we should be moving to a more mature configuration package. Something like [viper](https://github.com/spf13/viper). I've used it in the past to deal to good success. It has facilities to tie in environment variables with config files, and also gives us a path forward towards doing on-the-fly swapping.

OMG, i forgot the Viper, it's even can watch configure's change in runtime!

andresuribe87:
> > > IMO, we should be moving to a more mature configuration package. Something like [viper](https://github.com/spf13/viper). I've used it in the past to deal to good success. It has facilities to tie in environment variables with config files, and also gives us a path forward towards doing on-the-fly swapping.
> > 
> > 
> > OMG, i forgot the Viper, it's even can watch configure's change in runtime!
> 
> @andresuribe87 After some test and researches. `viper` do use `go-toml`, but will lose `comment` feature due the implementation and `watch` on mac will have some problems. I will prefer to keep this solution until we find a better configuration package.

I'm not familiar with the `comment` feature. Is it something that we need?

Re: the `watch` feature, would it make sense to fix those problems in the package? I understand wanting to keep this simple, but I think this is the perfect chance to find a good config package :)

andresuribe87:
Thanks @suutaku , appreciate the concerns to make sure the configuration we end up with is robust. See some comments inline below.

> we need `comment` so that devs will know what configure item means. about the configure features i think we can make a check list and discuss. maybe i missed some features, if need please fill it.

Do you mean the ability to add a `comment` in the configuration file? This should be supported by TOML. Are you saying that viper wouldn't be able to handle TOML files that have comments in them?

Regarding the requirements:
* I think the `watch` is a nice, but making configuration aspects work dynamically is not trivial at all. Hence I would suggest that this shouldn't be requirement.
* I don't fully follow what you mean by `command line binding`. Do you mean specifying flags? 
* I would argue that "environment variable" isn't really needed. Instead, we should pass in the path of the configuration file as a flag.

One additional requirement (and we're trying to apply this to every PR) is to minimize the new dependencies introduced. If we decide to introduce one, it should be a package that's actively maintained and mature enough to reduce supply chain security risks. Because of this, I would like to use a config package that's well known (hence the suggestion to use viper).

suutaku:
@andresuribe87, thanks to reply, about the questions:

> Do you mean the ability to add a comment in the configuration file? This should be supported by TOML

yes. `go-toml` supported comment and viper use `go-toml` as `toml `encoder/decoder

> Are you saying that viper wouldn't be able to handle TOML files that have comments in them?

yes.  `viper` load any type of struct to a map[string]interface{}. [code](https://github.com/spf13/viper/blob/ab1367a4d9f7dcad1b8ce840862b11e7d1f8f1ab/viper.go#L1544), that make comments  lose when overwrite exist configure file or create a new configure file.

I'm agree for not using the `command line binding` function. Because its effect is the same as loading the configuration file directly.
For the use of environment variables (a home directory), I mainly consider putting files such as databases and logs in a unified place.
I think `viper` is a usable solution, but it is not perfect.
I suggest removing the currently used third-party package and writing a new code at the current repo?

andresuribe87:
> I suggest removing the currently used third-party package and writing a new code at the current repo?

That makes sense to me!

> > Are you saying that viper wouldn't be able to handle TOML files that have comments in them?
>
>yes. viper load any type of struct to a map[string]interface{}. [code](https://github.com/spf13/viper/blob/ab1367a4d9f7dcad1b8ce840862b11e7d1f8f1ab/viper.go#L1544), that make comments lose when overwrite exist configure file or create a new configure file.

I think I get it now! What you're suggesting we should have is something similar to a schema for the config file, that describes what each possible parameter is, and what it does. Is that right? If so, what do you think about unmarshalling the config into the struct you created above according to [this](https://github.com/spf13/viper/#unmarshaling)?

suutaku:
@andresuribe87 
> what do you think about unmarshalling the config into the struct you created above according to [this](https://github.com/spf13/viper/#unmarshaling)? 

as long as we call `WriteConfig` or `WriteConfigAs`, comments lost. as you described unmashal `map[string]interface{}` to our's struct, marshal our's struct to `[]bytes` and write done. the comments information can be keepped. In such a case, it seems unnecessary for us to use `viper`, just write some simple code to `config/`.

consider remove the 3rd-party package,  if you guys agree to write a simple configure parser in `config/`, i will close this PR and create new one.

cmd/main.go
decentralgabe: i think it's important to take a hybrid approach
meaning we should define a precedence and overrides
1. default config
2. config file
3. env vars
what do you think?

pkg/storage/bolt.go
```diff
@@ -30,7 +35,14 @@ func (b *BoltDB) Init(options interface{}) error {
 	if b.db != nil && b.IsOpen() {
 		return fmt.Errorf("bolit db already opened with name %s", b.URI())
 	}
-	dbFilePath := fmt.Sprintf("%s_%s.db", DBFilePrefix, b.Type())
+	_, err := os.Stat(DBRootPath)
+	if os.IsNotExist(err) {
+		err = os.MkdirAll(DBRootPath, os.ModePerm)
+		if err != nil {
+			panic(err)
```
decentralgabe, line 42: let's return an error

pkg/storage/bolt.go
```diff
@@ -30,7 +35,14 @@ func (b *BoltDB) Init(options interface{}) error {
...
+		}
+	}
+	dbFilePath := fmt.Sprintf("%s/%s_%s.db", DBRootPath, DBFilePrefix, b.Type())
```
decentralgabe, line 45: should this be passed in directly?

pkg/storage/bolt.go
```diff
@@ -30,7 +35,14 @@ func (b *BoltDB) Init(options interface{}) error {
 	if b.db != nil && b.IsOpen() {
 		return fmt.Errorf("bolit db already opened with name %s", b.URI())
```
decentralgabe, line 36: not yours but how about changing to say `bolt db ...`

PR #229, opened 2022-12-20
[OSE-292] SubmitApplication now returns an Operation

andresuribe87:
# Overview
A credential application needs to be reviewed (by an automated or manual process). This asynchronous nature is better suited to be a long running operation.

# Description
This is part of the changes needed to support manual and automated issuance of credentials after an application is submitted.
Some of the changes include moving files into their own packages. This was done so break dependency cycles.

# How Has This Been Tested?
This is a refactor. Tests still pass (with some slight modifications)

# Checklist

## References
SIP5


pkg/server/router/operation.go
```diff
@@ -188,3 +201,22 @@ func (o OperationRouter) CancelOperation(ctx context.Context, w http.ResponseWri
...
+	Result OperationResult `json:"result,omitempty"`
+}
+
+type OperationResult struct {
+	// Populated when there was an error with the operation.
+	Error string `json:"error,omitempty"`
+
+	// Populated iff Error == "". The type should be specified in the calling APIs documentation.
+	Response any `json:"response,omitempty"`
+}
```
andresuribe87, line 222: This was moved from `pkg/server/router/presentation.go`

pkg/service/operation/credential/response.go
```diff
@@ -0,0 +1,35 @@
+package credential
```
nitro-neal, line 1: this file is basically the same as the presentation/response file right? (except the parentResource)

Is there anyway we can combine, or do you want it to be separate so they can implement different status if they want to

pkg/server/router/operation.go
```diff
@@ -188,3 +201,22 @@ func (o OperationRouter) CancelOperation(ctx context.Context, w http.ResponseWri
...
 }
+
+type Operation struct {
+	// The name of the resource related to this operation. E.g. "presentations/submissions/<uuid>"
+	ID string `json:"id"`
+
+	// Whether this operation has finished.
+	Done bool `json:"done"`
+
+	// Populated Done == true.
```
nitro-neal, line 212: // Populated if Done == true.

pkg/server/router/manifest.go
```diff
@@ -313,20 +314,14 @@ func (mr ManifestRouter) SubmitApplication(ctx context.Context, w http.ResponseW
...
 	}
 
-	resp := SubmitApplicationResponse{
-		Response:    submitApplicationResponse.Response,
-		Credentials: submitApplicationResponse.Credentials,
-		ResponseJWT: submitApplicationResponse.ResponseJWT,
-	}
-
-	return framework.Respond(ctx, w, resp, http.StatusCreated)
+	return framework.Respond(ctx, w, routerModel(*op), http.StatusCreated)
```
nitro-neal, line 324: eventually we can create a routerModel everywhere

nitro-neal (approved):
Looks great!

It seems like steelthread integration test still works with minimal changes.

In the future we should make another steelthread test that tests the asynchronous flow when it is ready

pkg/server/router/manifest.go
andresuribe87: For `Operation` objects, yes!

pkg/server/router/operation.go
andresuribe87: Done.

pkg/service/operation/credential/response.go
andresuribe87: Status is already slightly different: rejected vs denied, fulfilled vs approved. 

I think there is a chance of unifying, but I'm not comfortable yet deciding what the correct generalization would look like.

andresuribe87 (commented):
> In the future we should make another steelthread test that tests the asynchronous flow when it is ready

💯

PR #230, opened 2022-12-21
[OSE-156] Scaffolding and API for issuance templates.

andresuribe87:
# Overview
This PR adds the scaffolding and service APIs needed for issuance templates.

# Description
What's included in this PR.
* Router layer for Create/Read/Delete/List Issuance Templates.
* Service layer for the same operations. 
* Service layer models that describe the request/response objects

# How Has This Been Tested?
Existing tests pass. Unit tests shall be added when the implementation is done.

# Checklist

- [ ] I have read the CONTRIBUTING document.
- [ ] My code is consistent with the rest of the project 
- [ ] I have tagged the relevant reviewers and/or interested parties
- [ ] I have updated the READMEs and other documentation of affected packages

## References
SIP5 - see issuance template section.

### DIFF
--- a/pkg/server/router/issuance.go
+++ b/pkg/server/router/issuance.go
@@ -53,7 +53,7 @@ func (ir IssuanceRouter) CreateIssuanceTemplate(ctx context.Context, w http.Resp
 // DeleteIssuanceTemplate godoc
 // @Summary      Delete issuance template
 // @Description  Delete issuance template by ID
-// @Tags         IssuanceAPI
+// @Tags         IssuingAPI
 // @Accept       json
 // @Produce      json
 // @Param        id   path      string  true  "ID"
@@ -72,7 +72,7 @@ type ListIssuanceTemplatesResponse struct {
 // ListIssuanceTemplates godoc
 // @Summary      Lists issuance templates
 // @Description  Lists all issuangce templates stored in this service.
-// @Tags         IssuanceAPI
+// @Tags         IssuingAPI
 // @Accept       json
 // @Produce      json
 // @Success      200      {object}  ListIssuanceTemplatesResponse
--- a/pkg/server/router/issuance.go
+++ b/pkg/server/router/issuance.go
@@ -1,0 +1,84 @@
+package router
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/pkg/errors"
+	"github.com/tbd54566975/ssi-service/pkg/service/framework"
+	"github.com/tbd54566975/ssi-service/pkg/service/issuing"
+)
+
+type IssuanceRouter struct {
+	service issuing.Service
+}
+
+func NewIssuanceRouter(svc framework.Service) (*IssuanceRouter, error) {
+	service, ok := svc.(*issuing.Service)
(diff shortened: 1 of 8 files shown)


pkg/server/router/issuance.go
```diff
@@ -0,0 +1,84 @@
...
+// @Summary      Lists issuance templates
+// @Description  Lists all issuangce templates stored in this service.
+// @Tags         IssuingAPI
+// @Accept       json
+// @Produce      json
+// @Success      200      {object}  ListIssuanceTemplatesResponse
+// @Failure      400      {string}  string  "Bad request"
+// @Failure      500      {string}  string  "Internal server error"
+// @Router       /v1/manifests [get]
+func (ir IssuanceRouter) ListIssuanceTemplates(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
```
nitro-neal, line 82: In our other routes do we have it called Get instead of List?

pkg/service/issuing/model.go
```diff
@@ -0,0 +1,149 @@
...
+	// Info required to create a credential from a credential application.
+	Credentials []CredentialTemplate `json:"credentials"`
+}
+
+type GetIssuanceTemplateResponse struct {
+	// The template that was requested.
+	IssuanceTemplate IssuanceTemplate `json:"issuanceTemplate"`
+}
+
+type CreateIssuanceTemplateRequest struct {
```
nitro-neal, line 73: In most other places we don't let the user set the ID, we create a new ID with the object newly created object.

Is the flow for this needed where the user needs to create a unique id?

Do we have plans on how to handle it if the id is already taken

pkg/service/issuing/model.go
```diff
@@ -0,0 +1,149 @@
...
+
+func (s *Service) CreateIssuanceTemplate(request *CreateIssuanceTemplateRequest) (*IssuanceTemplate, error) {
+	return nil, nil
+}
+
+func (s *Service) DeleteIssuanceTemplate(request *DeleteIssuanceTemplateRequest) error {
+	return nil
+}
+
+func (s *Service) ListIssuanceTemplates(request *ListIssuanceTemplatesRequest) (*ListIssuanceTemplatesResponse, error) {
```
nitro-neal, line 147: We have this as Get in other places, I do kind of like List better.

Eventually we can change everything to be List with a filter if it makes sense

nitro-neal (approved):
Looks great!

pkg/server/router/issuance.go
andresuribe87: I'm partial to `List`. I think we currently have both :(

I can take a pass at unifying in a future PR!

pkg/service/issuing/model.go
andresuribe87: Excellent point!

I made some changes so that the ID is not required for the endpoint (this is the router one), but still is required in the service.

pkg/service/issuing/model.go
andresuribe87: Will normalize in a future PR.

PR #234, opened 2023-01-03
[OSE-300] Credential applications can be cancelled

andresuribe87:
# Overview
After a credential application is submitted, an operation is returned. This PR adds the ability to cancel the operation related to that application.

# Description
All operations can be cancelled, as long as they aren't marked as done.

# How Has This Been Tested?
Added a simple unit test to the service endpoint.

# Checklist


pkg/service/manifest/storage/storage.go
```diff
@@ -30,6 +31,8 @@ type StoredManifest struct {
 
 type StoredApplication struct {
 	ID             string                         `json:"id"`
+	Status         credential.Status              `json:"status"`
+	Reason         string                         `json:"reason"`
```
decentralgabe, line 35: should this be an enum or user defined?

decentralgabe (approved):

pkg/service/manifest/storage/storage.go
andresuribe87: I'm imagining it's user defined free form text. We can change later once we see real use cases.

PR #236, opened 2023-01-04
adding env var handling

nitro-neal:
# Overview
Adding environment variables that get loading into the ssi-service config on startup

# Description
This change refactors the LoadConfig and adds the mutation of the config in seperate functions.
checkValidConfigPath - This checks if a config path is provided and if a valid config is at the path
parseAndApplyDefaults - Creates a new config object and assigns the base defaults
loadDefaultServicesConfig - If there is no toml file we will programmatically create the config
loadTomlFileConfig - if there is a toml file present we will load in the toml config
applyEnvVariables - finally after config is loaded if there are environment variables we will add these to the config


config/config.go
```diff
@@ -9,17 +9,21 @@ import (
...
 	"github.com/BurntSushi/toml"
 	"github.com/ardanlabs/conf"
+	"github.com/joho/godotenv"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
 
 const (
 	DefaultConfigPath = "config/config.toml"
+	DefaultEnvPath    = "config/.env"
```
decentralgabe, line 19: does this pick up every `*.env` file in the directory?

config/config.go
```diff
@@ -9,17 +9,21 @@ import (
...
 	ConfigFileName    = "config.toml"
 	ServiceName       = "ssi-service"
 	ConfigExtension   = ".toml"
 
-	DefaultServiceEndpoint = "http://localhost:8080"
+	DefaultServiceEndpoint      = "http://localhost:8080"
+	EnvVariableKeystorePassword = "KEYSTORE_PASSWORD"
```
decentralgabe, line 25: may want to make a `type EnvironmentVariable string` for these

config/config.go
```diff
@@ -155,84 +159,155 @@ func (p *PresentationServiceConfig) IsEmpty() bool {
...
+	}
+
+	// create the config object
+	var config SSIServiceConfig
+
+	if err := parseAndApplyDefaults(config); err != nil {
+		return nil, errors.Wrap(err, "parse and apply defaults")
+	}
+
+	if loadDefaultConfig == true {
```
decentralgabe, line 174:
```suggestion
	if loadDefaultConfig {
```

config/config.go
```diff
@@ -155,84 +159,155 @@ func (p *PresentationServiceConfig) IsEmpty() bool {
...
+		loadDefaultServicesConfig(&config)
+	} else {
+		if err := loadTomlFileConfig(path, &config); err != nil {
+			return nil, errors.Wrap(err, "load toml config ")
```
decentralgabe, line 178: extra space

config/config.go
```diff
@@ -155,84 +159,155 @@ func (p *PresentationServiceConfig) IsEmpty() bool {
...
+		},
+		IssuingServiceConfig: IssuingServiceConfig{
+			BaseServiceConfig: &BaseServiceConfig{Name: "issuing"},
+		},
+	}
+
+	config.Services = servicesConfig
+}
+
+func loadTomlFileConfig(path string, config *SSIServiceConfig) error {
```
decentralgabe, line 264:
```suggestion
func loadTOMLConfig(path string, config *SSIServiceConfig) error {
```

config/config.go
```diff
@@ -155,84 +159,155 @@ func (p *PresentationServiceConfig) IsEmpty() bool {
...
+	if present {
+		if config.Services.StorageOption == nil {
+			config.Services.StorageOption = make(map[string]interface{})
 		}
 
-		// apply defaults if not included in toml file
-		if config.Services.CredentialConfig.BaseServiceConfig.ServiceEndpoint == "" {
-			config.Services.CredentialConfig.BaseServiceConfig.ServiceEndpoint = config.Services.ServiceEndpoint
+		storageOptionMap, ok := config.Services.StorageOption.(map[string]interface{})
+		if ok == false {
```
decentralgabe, line 304:
```suggestion
		if !ok {
```

decentralgabe (commented):
looks good - small comments. may be worth adding a section to the readme on config and the order in which it is loaded

config/config.go
andresuribe87, line 304:
```suggestion
		if !ok {
```

config/config.go
```diff
@@ -155,84 +159,155 @@ func (p *PresentationServiceConfig) IsEmpty() bool {
...
+			return errors.New("storage option casting")
```
andresuribe87, line 305:
```suggestion
			return errors.New("storage option must be of type map[string]interface{}")
```

config/.env.example
```diff
@@ -0,0 +1,2 @@
+KEYSTORE_PASSWORD=default-keystore-password
+DB_PASSWORD=default-db-password
```
andresuribe87, line 2: This is awesome! Would you mind adding instructions to the README?

andresuribe87 (approved):
Looks good! Couple small suggestions.

config/config.go
nitro-neal: no, just a file named .env in the config/ dir

config/config.go
nitro-neal: Yea that makes sense

config/.env.example
nitro-neal: yea good call

PR #238, opened 2023-01-05
fixing streamlit

nitro-neal:
- Adding streamlit to networks to fix it
- Creating default schema creation so that we can issue a VC with streamlit


nitro-neal:
image

andresuribe87 (approved):
This is sick!

PR #239, opened 2023-01-06
[OSE-311] Docs publish

michaelneale:
Trying this from a branch - same as https://github.com/TBD54566975/ssi-service/pull/198


andresuribe87 (approved):
Thank you!

andresuribe87:
A couple of issues that need to be addressed before merging this PR:
* Automatic commits do not trigger github actions. More details about this in this [README](https://github.com/stefanzweifel/git-auto-commit-action#commits-made-by-this-action-do-not-trigger-new-workflow-runs), and potential solutions (like creating a robot account with a personal access token).
* Code generation is not idempotent (try running `mage spec`, then `git commit -a -m "commiting spec"`, then `mage spec`; you'll notice that there is still a diff when you run `git diff --shortstat`). This would cause and inifinite recursion of github actions if we were to trigger actions on automatic commits.

michaelneale:
@andresuribe87 yeah I guess it’s good it doesn’t trigger actions though right? (That must be by design?). But the statuses end up hanging. I wonder if there is a way to make things green when it’s from the action bot.

michaelneale:
Maybe can do it just on merge to main? Not on each PR?

michaelneale:
hrm - the conflict will be annoying. 

I wonder if there is a better way to publish docs - should they be stored in the repo? (I guess so).

decentralgabe (approved):

decentralgabe:
> Maybe can do it just on merge to main? Not on each PR?

it should be done on each commit to avoid it getting out of sync

michaelneale:
@andresuribe87 you want to merge this one or the other one?

andresuribe87:
> @andresuribe87 you want to merge this one or the other one?

Unfortunately, we cannot merge this yet until we figure out the issues mentioned. I'll be digging into it further this week.

decentralgabe:
a dumb approach, that might just work is to have a check that fails if the swagger is not up to date. this forces the user to regen and re-commit and we don't need to worry bout the automation headache

andresuribe87:
Note that swaggo currently is not idempotent. In particular, the approach of generating the file and detecting a diff to see if swagger is up to date would not work since as the result will always yield a diff. 

We could do a hack where we measure the diff size. I'm not sure if that's stable though.

decentralgabe:
how do we make it idempotent?

michaelneale:
closing as this was just to try something out.

PR #240, opened 2023-01-06
ReviewApplication endpoint for fulfilling or denying applications

andresuribe87:
# Overview
The existing SubmitApplication endpoint was updated so it returns an Operation that is not done. In order to mark the operation as done, it needs to be cancelled, or the application needs to be reviewed. This PR does the latter. Note that when an application is reviewed, the application's status is updated and a response is also recorded in the DB. 

# How Has This Been Tested?
As this doesn't introduce new functionality, the existing tests suffice.

# Tracking
OSE-203


pkg/service/manifest/response.go
```diff
@@ -80,8 +81,14 @@ func (s Service) buildCredentialResponse(applicantDID, manifestID, applicationID
 	}
 
 	// set the information for the fulfilled credentials in the response
-	if err := responseBuilder.SetFulfillment(descriptors); err != nil {
-		return nil, nil, util.LoggingErrorMsg(err, "could not fulfill credential application: could not set fulfillment")
+	if approved {
+		if err := responseBuilder.SetFulfillment(descriptors); err != nil {
+			return nil, nil, util.LoggingErrorMsg(err, "could not fulfill credential application: could not set fulfillment")
+		}
+	} else {
```
decentralgabe, line 88: else is unnecessary given the return?

pkg/service/manifest/storage/storage.go
```diff
@@ -237,3 +238,54 @@ func (ms *Storage) DeleteResponse(id string) error {
...
+//     creates in step 2.
+//
+// The operation and it's response (from 3) are returned.
+func (ms *Storage) ReviewApplication(id string, approved bool, reason string, opID string, response StoredResponse) (*StoredResponse, *opstorage.StoredOperation, error) {
+	// TODO: everything should be in a single Tx.
+	m := map[string]any{
+		"status": opsubmission.StatusDenied,
+		"reason": reason,
+	}
+	if approved {
```
decentralgabe, line 255: concurreny access risk?

thereisnogabe (approved):

decentralgabe:
noticed the e2e test failing - is that new? cc: @nitro-neal

pkg/service/manifest/response.go
andresuribe87: It's necessary. Only on of  `fulfillment` or `denial` can be set.

pkg/service/manifest/storage/storage.go
andresuribe87: Yes! This I'm leaving that for a separate PR. I wanted to get a simple implementation out, and then fix the concurrent access issues.

andresuribe87 (commented):
> noticed the e2e test failing - is that new? cc: @nitro-neal

Failure was introduced in this PR. Fixed it now. Yay for tests!

integration/didweb_integration_test.go
```diff
@@ -118,7 +119,7 @@ func TestDidWebCreateCredentialManifestIntegration(t *testing.T) {
 	assert.NotEmpty(t, manifestID)
 }
 
-func TestDidWebSubmitApplicationIntegration(t *testing.T) {
+func TestDidWebSubmitAndReviewApplicationIntegration(t *testing.T) {
```
decentralgabe, line 122: for the next one `TestDIDWeb`... I know you didn't introduce this

Issue #176, opened 2022-11-14
Create namespaces on db creation


nitro-neal:
Currently we do an empty write to the credential bolt db to get around a bug where if you do a read before anything is written it will throw an error

it's because the namespace bolt is using is created on first write. we should open up another issue to address it.
solutions:

seed the DB with a first entry
detect the namespace not exists error and throw another error

andresuribe87:
@nitro-neal is this still an issue?

PR #241, opened 2023-01-06
fixing namespace seeding issue

nitro-neal:
Currently we do an empty write to the credential bolt db to get around a bug where if you do a read before anything is written it will throw an error

it's because the namespace bolt is using is created on first write. we should open up another issue to address it.
solutions:

https://github.com/TBD54566975/ssi-service/issues/176

I have handled this error by logic in the code.

If we think a better solution would be to NOT return an error if the bucket / namespace does not exist I am happy to implement that fix instead of this.

### DIFF
--- a/pkg/service/credential/service.go
+++ b/pkg/service/credential/service.go
@@ -220,7 +220,7 @@ func (s Service) CreateCredential(request CreateCredentialRequest) (*CreateCrede
 func getStatusListCredential(s Service, issuerID string, schemaID string) (*credential.VerifiableCredential, error) {
 	storedStatusListCreds, err := s.storage.GetStatusListCredentialsByIssuerAndSchema(issuerID, schemaID)
 	if err != nil {
-		return nil, util.LoggingNewErrorf("problem with getting status list credential for issuer: %s schema: %s", issuerID, schemaID)
+		logrus.Warnf("problem with getting status list credential for issuer: %s schema: %s. Continuing because this could be first time accessing this pair...", issuerID, schemaID)
 	}
 
 	// This should never happen, there should always be only 1 status list credential per <issuer,schema> pair
--- a/pkg/service/credential/storage.go
+++ b/pkg/service/credential/storage.go
@@ -78,25 +78,6 @@ func NewCredentialStorage(db storage.ServiceStorage) (*Storage, error) {
 		return nil, errors.New("bolt db reference is nil")
 	}
 
-	// TODO: (Neal) there is a current bug with our Bolt implementation where if we do a GET without anything in the db it will throw an error
-	// Doing initial writes and then deleting will "warm up" our database and when we do a GET after that it will not crash and return empty list
-	// https://github.com/TBD54566975/ssi-service/issues/176
-	if err := db.Write(credentialNamespace, fakeKey, nil); err != nil {
-		return nil, util.LoggingErrorMsg(err, "problem writing status initial write to db")
-
-	}
-	if err := db.Delete(credentialNamespace, fakeKey); err != nil {
-		return nil, util.LoggingErrorMsg(err, "problem with initial delete to db")
-	}
-
-	if err := db.Write(statusListCredentialNamespace, fakeKey, nil); err != nil {
-		return nil, util.LoggingErrorMsg(err, "problem writing status initial write to db")
-	}
-
-	if err := db.Delete(statusListCredentialNamespace, fakeKey); err != nil {
-		return nil, util.LoggingErrorMsg(err, "problem with initial delete to db")
-	}
-
 	randUniqueList := randomUniqueNum(bitStringLength)
 	uniqueNumBytes, err := json.Marshal(randUniqueList)
 	if err != nil {


pkg/service/credential/service.go
```diff
@@ -220,7 +220,7 @@ func (s Service) CreateCredential(request CreateCredentialRequest) (*CreateCrede
 func getStatusListCredential(s Service, issuerID string, schemaID string) (*credential.VerifiableCredential, error) {
 	storedStatusListCreds, err := s.storage.GetStatusListCredentialsByIssuerAndSchema(issuerID, schemaID)
 	if err != nil {
-		return nil, util.LoggingNewErrorf("problem with getting status list credential for issuer: %s schema: %s", issuerID, schemaID)
+		logrus.Warnf("problem with getting status list credential for issuer: %s schema: %s. Continuing because this could be first time accessing this pair...", issuerID, schemaID)
```
decentralgabe, line 223: `logrus.WithError(err).Warnf(...`

decentralgabe (changes requested):
I found the bug.

If you see [ReadPrefix](https://github.com/TBD54566975/ssi-service/blob/main/pkg/storage/bolt.go#L108) and [ReadAllKeys](https://github.com/TBD54566975/ssi-service/blob/main/pkg/storage/bolt.go#L144) we return an error; however, in [Read](https://github.com/TBD54566975/ssi-service/blob/main/pkg/storage/bolt.go#L94) and in [ReadAll](https://github.com/TBD54566975/ssi-service/blob/main/pkg/storage/bolt.go#L128) we do not.

I think all of these should not return an error and instead return a warning.

Similarly, for deletes [we may want to](https://github.com/TBD54566975/ssi-service/blob/main/pkg/storage/bolt.go#L159) change to a warning as well.

decentralgabe:
> I found the bug.
> 
> 
> 
> If you see [ReadPrefix](https://github.com/TBD54566975/ssi-service/blob/main/pkg/storage/bolt.go#L108) and [ReadAllKeys](https://github.com/TBD54566975/ssi-service/blob/main/pkg/storage/bolt.go#L144) we return an error; however, in [Read](https://github.com/TBD54566975/ssi-service/blob/main/pkg/storage/bolt.go#L94) and in [ReadAll](https://github.com/TBD54566975/ssi-service/blob/main/pkg/storage/bolt.go#L128) we do not.
> 
> 
> 
> I think all of these should not return an error and instead return a warning.
> 
> 
> 
> Similarly, for deletes [we may want to](https://github.com/TBD54566975/ssi-service/blob/main/pkg/storage/bolt.go#L159) change to a warning as well.

Neal I see you mentioned this too. Sorry, missed that.

andresuribe87:
> If we think a better solution would be to NOT return an error if the bucket / namespace does not exist I am happy to implement that fix instead of this.

Heavy +1 to this.

pkg/service/credential/storage.go
```diff
@@ -78,25 +78,6 @@ func NewCredentialStorage(db storage.ServiceStorage) (*Storage, error) {
...
-	}
-
-	if err := db.Write(statusListCredentialNamespace, fakeKey, nil); err != nil {
-		return nil, util.LoggingErrorMsg(err, "problem writing status initial write to db")
-	}
-
-	if err := db.Delete(statusListCredentialNamespace, fakeKey); err != nil {
-		return nil, util.LoggingErrorMsg(err, "problem with initial delete to db")
-	}
-
```
andresuribe87, line 99: 😍

pkg/storage/bolt.go
```diff
@@ -105,9 +105,8 @@ func (b *BoltDB) ReadPrefix(namespace, prefix string) (map[string][]byte, error)
 	err := b.db.View(func(tx *bolt.Tx) error {
 		bucket := tx.Bucket([]byte(namespace))
 		if bucket == nil {
-			errMsg := fmt.Sprintf("namespace<%s> does not exist", namespace)
-			logrus.Error(errMsg)
-			return errors.New(errMsg)
+			logrus.Infof("namespace<%s> does not exist", namespace)
```
andresuribe87, line 108: I don't think we should log on this since it's expected behavior.

Instead, a test can detect whether the implementation is correct. Is there a unit test covering this behavior?

pkg/storage/bolt.go
```diff
@@ -141,7 +140,8 @@ func (b *BoltDB) ReadAllKeys(namespace string) ([]string, error) {
 	err := b.db.View(func(tx *bolt.Tx) error {
 		bucket := tx.Bucket([]byte(namespace))
 		if bucket == nil {
-			return util.LoggingNewErrorf("namespace<%s> does not exist", namespace)
+			logrus.Infof("namespace<%s> does not exist", namespace)
+			return nil
```
andresuribe87, line 144: Same comment as above.

pkg/storage/bolt.go
nitro-neal: Yup agree, 

I have created unit tests and removed this log

pkg/storage/bolt.go
```diff
@@ -141,7 +136,7 @@ func (b *BoltDB) ReadAllKeys(namespace string) ([]string, error) {
 	err := b.db.View(func(tx *bolt.Tx) error {
 		bucket := tx.Bucket([]byte(namespace))
 		if bucket == nil {
-			return util.LoggingNewErrorf("namespace<%s> does not exist", namespace)
+			return nil
```
decentralgabe, line 139: I would keep the error as a warn or info

decentralgabe (approved):

PR #246, opened 2023-01-09
ctx for db write

nitro-neal:
# Overview
Adding context to database impl

# Description
This change adds the api level context and passes wires it up to the db Write functionality. 

The context object from the http request includes information about the client making the request and will end if the client disconnects before the request is finished.


decentralgabe:
looks like a few more method calls need ctx added as arguments

nitro-neal:
yea I should have put this up as a draft, just did a write to make sure this was the way we wanted to do it.

nitro-neal:
will implement the rest of the db interface methods now and update unit tests

andresuribe87 (approved):

PR #247, opened 2023-01-11
Automatic Credentials based on Issuance Tempaltes

andresuribe87:
# Overview
This PR introduces issuance of credentials according to a stored issuance template.

# Description
When an issuance template exists in storage, we are able to issue credentials automatically. This PR introduces that automatic step. It's done immediately after the existing submission logic finishes.

# How Has This Been Tested?
A new test was added where an issuance template exists for the given application, and credentials were verified to have been created with the correct values.

## References
This is part of SIP 5


pkg/server/server_manifest_test.go
```diff
@@ -308,6 +314,152 @@ func TestManifestAPI(t *testing.T) {
...
+		applicantPrivKeyBytes, err := base58.Decode(applicantDID.PrivateKeyBase58)
+		assert.NoError(tt, err)
+		applicantPrivKey, err := crypto.BytesToPrivKey(applicantPrivKeyBytes, applicantDID.KeyType)
+		assert.NoError(tt, err)
+		signer, err := keyaccess.NewJWKKeyAccess(applicantDID.DID.ID, applicantPrivKey)
+		assert.NoError(tt, err)
+		signed, err := signer.SignJSON(applicationRequest)
+		assert.NoError(tt, err)
+
+		now := time.Date(2022, 10, 31, 0, 0, 0, 0, time.UTC)
```
decentralgabe, line 410: nit: this is not now 😄

pkg/server/server_manifest_test.go
```diff
@@ -308,6 +314,152 @@ func TestManifestAPI(t *testing.T) {
...
+		assert.True(tt, op.Done)
+
+		var appResp router.SubmitApplicationResponse
+		respData, err := json.Marshal(op.Result.Response)
+		assert.NoError(tt, err)
+		err = json.Unmarshal(respData, &appResp)
+		assert.NoError(tt, err)
+
+		assert.Len(tt, appResp.Credentials, 2, "each output_descriptor in the definition should result in a credential")
+		vc, err := signing.ParseVerifiableCredentialFromJWT(appResp.Credentials.([]any)[0].(string))
```
decentralgabe, line 435: should this method be renamed to reflect it can parse multiple credentials?

pkg/service/manifest/response.go
```diff
@@ -39,13 +47,34 @@ func (s Service) signCredentialResponseJWT(ctx context.Context, signingDID strin
...
+		return nil, nil, util.LoggingErrorMsgf(
+			err,
+			"could not fulfill credential credentials: could not set credentials id: %s",
+			applicationID,
+		)
 	}
 
+	templateMap := make(map[string]*issuing.CredentialTemplate)
+	if template != nil {
+		for _, v := range template.Credentials {
```
decentralgabe, line 73: nit: meaningful variable instead of `v` like `credential` or `cred`

pkg/service/manifest/response.go
```diff
@@ -98,7 +146,107 @@ func (s Service) buildCredentialResponse(ctx context.Context, applicantDID, mani
...
+	ct, ok := templateMap[od.ID]
+	if !ok {
+		logrus.Warnf("Did not find output_descriptor with ID \"%s\" in template. Skipping application.", od.ID)
+		return nil
+	}
+	c, err := getCredential(applicationJSON, ct, credManifest, application.PresentationSubmission)
+	if err != nil {
+		return err
+	}
+	for k, v := range ct.Data.Claims.Data {
```
decentralgabe, line 169: is data guaranteed to not be nil, same with claims and data?

pkg/service/manifest/response.go
```diff
@@ -98,7 +146,107 @@ func (s Service) buildCredentialResponse(ctx context.Context, applicantDID, mani
...
+		claimValue := v
+		vs, ok := v.(string)
+		if ok {
```
decentralgabe, line 172: this can be one line

pkg/service/manifest/response.go
```diff
@@ -98,7 +146,107 @@ func (s Service) buildCredentialResponse(ctx context.Context, applicantDID, mani
...
+	return nil
+}
+
+func getCredential(
+	applicationJSON map[string]any,
+	ct *issuing.CredentialTemplate,
+	credManifest manifest.CredentialManifest,
+	submission *exchange.PresentationSubmission,
+) (any, error) {
+	if ct.Data.CredentialInputDescriptor != "" {
```
decentralgabe, line 201: consider inverting to reduce nesting

```go
if ct.Data.CredentialInputDescriptor == "" {
   // maybe log that this is empty
   return nil, nil
}
if credManifest.PresentationDefinition == nil {
 ...
}
```

pkg/service/manifest/response.go
```diff
@@ -98,7 +146,107 @@ func (s Service) buildCredentialResponse(ctx context.Context, applicantDID, mani
...
+}
+
+func fromFormat(format exchange.JWTFormat, claim any) (any, error) {
+	switch format {
+	case exchange.JWTVC:
+		_, token, err := util.ParseJWT(keyaccess.JWT(claim.(string)))
+		if err != nil {
+			return nil, errors.Wrapf(err, "parsing jwt as %s", exchange.JWTVC)
+		}
+		return token.PrivateClaims()["vc"], nil
```
decentralgabe, line 240: might want to break this into two lines, there's a chance the `["vc"]` lookup can crash

pkg/service/manifest/service.go
```diff
@@ -286,13 +330,93 @@ func (s Service) ProcessApplicationSubmission(ctx context.Context, request model
...
+	if err != nil {
+		return nil, errors.Wrap(err, "fetching issuance templates by manifest ID")
+	}
+	if len(issuanceTemplates) == 0 {
+		return nil, nil
+	}
+
+	issuanceTemplate := issuanceTemplates[0].IssuanceTemplate
+
+	if len(issuanceTemplates) > 1 {
```
decentralgabe, line 369: maybe turn this into two checks

`if len(issuanceTemplates) > 0)` and then > 1

decentralgabe (approved):
nice works. lots of good stuff in here 🥳

pkg/server/server_manifest_test.go
andresuribe87: 🤣 

Injected this time into the service now for testing purposes.

pkg/service/manifest/service.go
decentralgabe: disregard this

PR #250, opened 2023-01-11
Added manual overrides for credential issuance upon application review.

andresuribe87:
# Overview
When a review happens, the issuer admin might want to override some values that will be put in the credentials that will be produced.

# Description
This PR changes. Noteworthy: using a map instead of a list for the overrides. This is because the order of the overrides doesn't matter.

# How Has This Been Tested?
Added overrides for a credential in existing tests. Additionally modified the e2e tests. 

## References
Still part of SIP5


pkg/service/manifest/response.go
```diff
@@ -98,6 +100,7 @@ func (s Service) buildCredentialResponse(
 				return nil, nil, err
 			}
 		}
+		s.applyRequestData(&credentialRequest, credentialOverrides, od)
```
decentralgabe, line 103: methods without return values make me nervous!

decentralgabe (approved):

PR #252, opened 2023-01-12
Batch Writes

nitro-neal:
# Overview
Transactional Writes for CreateCredential

# Description
This change does changes the way CreateCredential works in that it delays all writes until the end of the function and then transactionally writes them


nitro-neal:
This change essentially does all the writes needed for creating a VC at the end of the function in one call:
image

nitro-neal:
image

There is still a case where we can get in trouble. In this case the status list index would be 1 for both VCs (which is wrong)

decentralgabe (approved):

PR #255, opened 2023-01-13
Added e2e test for automatic issuance and bug fixes.

andresuribe87:
# Overview
Added the e2e for automatic issuance, which involves creating an IssuanceTemplate.

# Description
Some additional changes in this PR
* Logging the reason why a panic happened.
* Fixed a panic where the `err` was being returned though it was actually `nil`.
* Made the `GetResponse` return the same shape as what is in the `operation.result` of calling `SubmitApplication`.
* Slightly tweaked the issuance template shape so that `data` contains only the claims about credentialSubject.

# How Has This Been Tested?

The e2e test was added.
All other tests still pass.

## References
Final pieces of SIP 5.

### DIFF
--- a/integration/common.go
+++ b/integration/common.go
@@ -426,3 +426,23 @@ func ReviewApplication(params reviewApplicationParams) (string, error) {
 
 	return output, nil
 }
+
+type issuanceTemplateParams struct {
+	SchemaID   string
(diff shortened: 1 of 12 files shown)


decentralgabe (approved):
nice n clean

PR #261, opened 2023-01-19
Updated the Submission to contain the verifiablePresentation.

andresuribe87:
# Overview
Previously: a Submission object (which is returned with `GetSubmission`) contained metadata only.
Now: it contains the `VerifiablePresentation` object that was sent.

# Description
This is so API clients can retrieve what the information sent was. An follow up may be to include the JWT that was sent, instead of the actual decoded payload.

# How Has This Been Tested?
Unit

### DIFF
--- a/pkg/service/presentation/storage.go
+++ b/pkg/service/presentation/storage.go
@@ -140,7 +140,11 @@ func (ps *Storage) DeletePresentation(ctx context.Context, id string) error {
 }
 
 func (ps *Storage) StoreSubmission(ctx context.Context, s prestorage.StoredSubmission) error {
-	id := s.Submission.ID
+	sub, ok := s.VerifiablePresentation.PresentationSubmission.(exchange.PresentationSubmission)
+	if !ok {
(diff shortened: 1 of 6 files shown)


pkg/service/presentation/storage.go
```diff
@@ -140,7 +140,11 @@ func (ps *Storage) DeletePresentation(ctx context.Context, id string) error {
 }
 
 func (ps *Storage) StoreSubmission(ctx context.Context, s prestorage.StoredSubmission) error {
-	id := s.Submission.ID
+	sub, ok := s.VerifiablePresentation.PresentationSubmission.(exchange.PresentationSubmission)
+	if !ok {
```
nitro-neal, line 144: any pros and cons to doing it this way vs reflections vs json marshalling ?

nitro-neal (approved):
looks good!

pkg/service/presentation/storage.go
andresuribe87: Synced offline. Casting guarantees that correct type is the one present.

PR #263, opened 2023-01-23
Adding transactional scheme

nitro-neal:
# Overview
This change adds a transactional finality to our database for our CreateCredentials api endpoint.

# Description

We define the function for all the business logic. We pass the function down to the database layer. The database layer will call this function with the new Execute function to ensure that it is transactionally safe.

In our implementation in Redis the database layer calls the function but skips all writes. It does this in order to get the 'read-keys' so that we can add these keys to the redis watch call.

After we have the read keys we actually call the function for real, do the writes, and handle retry logic if a read key has changed during our write.

# Changes
- Adds Execute to our db interface to transactionally run business logic
- Adding redis as default backing db
- Adding redis-commander on port 9001 so we can browse database state
- Creates multithreading integration test
- Adds "github.com/cenkalti/backoff" dependency to implement exponential backoffs


pkg/storage/redis.go
```diff
@@ -62,9 +68,73 @@ func (b *RedisDB) Close() error {
 	return b.db.Close()
 }
 
+type TxContext struct {
+	tx        goredislib.Pipeliner
+	watchKeys []string
+	watchOnly bool
```
andresuribe87, line 74: Perhaps this is unused? I don't see where it's being set to true.

pkg/storage/redis.go
```diff
@@ -81,9 +151,19 @@ func (b *RedisDB) WriteMany(ctx context.Context, namespaces, keys []string, valu
...
-	res, err := b.db.Get(ctx, nameSpaceKey).Bytes()
+
+	if (*ctx).Value(TX) != nil {
+		txContext := (*ctx).Value(TX).(TxContext)
+
+		watchKeys := txContext.watchKeys
+		watchKeys = append(watchKeys, nameSpaceKey)
+
+		*ctx = context.WithValue(*ctx, TX, TxContext{txContext.tx, watchKeys, txContext.watchOnly})
+	}
```
andresuribe87, line 164: From the [godoc](https://pkg.go.dev/context), "Use context Values only for request-scoped data that transits processes and APIs, not for passing optional parameters to functions."

The recommendation from that doc is to follow those guidelines, which I agree with, so we get the benefits of static analysis.

I believe it's possible to do this by modifying the API. For example, you can have:
* A separate struct where you accumulate the things that need to be watched.
* A separate set of Read/Write functions to be used when you're doing transactions.

For the second part, I think it's ok (and even expected), that callers of the API should know what they're doing when executing transactions. 

Happy to take a stab at that together.

andresuribe87 (changes requested):

build/docker-compose.yml
```diff
@@ -30,7 +30,7 @@ services:
       context: ../
       dockerfile: build/Dockerfile-gui
     ports:
-      - "8081:8501"
+      - "8003:8501"
```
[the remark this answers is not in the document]
nitro-neal: swagger is on 8002 so I put our gui documentation thing on 8003.

And not to get confused with the 8081 that   redis-commander uses internally

integration/credential_revocation_concurrency_integration_test.go
```diff
@@ -0,0 +1,161 @@
...
+			assert.NoError(t, err)
+			assert.NotEmpty(t, credStatusListIndex)
+
+			credStatusListIndexes[i] = credStatusListIndex
+		}(i)
+	}
+
+	wg.Wait()
+
+	assert.True(t, areElementsUnique(credStatusListIndexes))
```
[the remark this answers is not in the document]
nitro-neal: I don't know what the indexes are because they are random, I would have to build a new route to get the db internals.
However if all of them are unique it means our transactional layer worked because there are no duplicates which was happening before

pkg/storage/redis.go
nitro-neal: Yea I'll change it up so that we have different read/write functions for tx. Was trying to force it in to be plug and play but I guess its ok if we have 2 different read/write

config/compose.toml
```diff
@@ -21,10 +21,14 @@ log_level = "info"
 
 enable_schema_caching = true
 
-# Bolt Configuration
+# Redis Configuration
```
decentralgabe, line 24: maybe just Storage Configuration?

pkg/service/credential/service.go
```diff
@@ -90,10 +90,30 @@ func NewCredentialService(config config.CredentialServiceConfig, s storage.Servi
...
+	return credResponse, nil
+}
+
+func (s Service) CreateCredentialFunc(request CreateCredentialRequest) storage.BusinessLogicFunc {
+	return func(ctx *context.Context) (any, error) {
+		return s.CreateCredentialBusinessLogic(ctx, request)
+	}
+}
+
+func (s Service) CreateCredentialBusinessLogic(ctx *context.Context, request CreateCredentialRequest) (*CreateCredentialResponse, error) {
```
decentralgabe, line 114: why does this accept a context pointer but other methods do not? let's be consistent

decentralgabe: to be clearer I don't think we should have ptrs

pkg/storage/redis.go
```diff
@@ -4,29 +4,35 @@ import (
...
 
+type ContextKey string
+
+type RedisDB struct {
+	db *goredislib.Client
+}
+
 func init() {
 	err := RegisterStorage(&RedisDB{})
 	if err != nil {
```
decentralgabe, line 31: not yours but nit
```suggestion
	if err := RegisterStorage(new(RedisDB));err != nil {
```

pkg/storage/redis.go
```diff
@@ -62,9 +68,73 @@ func (b *RedisDB) Close() error {
...
+	err = backoff.Retry(func() error {
+		err := b.db.Watch(ctx, txf, ctxWatchKeys...)
+		if err != nil && errors.Is(err, goredislib.TxFailedErr) {
+			logrus.Warn("Optimistic lock lost. Retrying..")
+		}
+		return err
+	}, expBackoff)
+
+	if err != nil {
+		logrus.Errorf("error after retrying: %v", err)
```
decentralgabe, line 118: should return `nil, err`?

pkg/storage/storage.go
```diff
@@ -26,14 +28,15 @@ type ServiceStorage interface {
 	Close() error
 	Write(ctx context.Context, namespace, key string, value []byte) error
 	WriteMany(ctx context.Context, namespace, key []string, value [][]byte) error
-	Read(ctx context.Context, namespace, key string) ([]byte, error)
+	Read(ctx *context.Context, namespace, key string) ([]byte, error)
```
decentralgabe, line 31: no pointer!

go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/alicebob/miniredis/v2 v2.30.0
 	github.com/ardanlabs/conf v1.5.0
 	github.com/benbjohnson/clock v1.3.0
+	github.com/cenkalti/backoff v2.2.1+incompatible
```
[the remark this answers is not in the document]
nitro-neal: Good catch

pkg/storage/redis.go
nitro-neal: Removed

pkg/storage/redis.go
nitro-neal: great catch yup

pkg/service/credential/service.go
nitro-neal: yup I was trying to overload this CTX object instead of making my own accumulator object, I have removed and no more ctx pointers

thereisnogabe (approved):

decentralgabe:
nice stuff!

nitro-neal:
Had a talk with @andresuribe87 and we came up with a solution that, instead of running the business logic in "read only" mode to get the watch keys, the plan is to make a change so that at the application layer you pass in the watch keys. 

This will ensure that what you want to be transactualy safe will be and we wont have to run the business logic twice and potentially get into a strange state. 👍

pkg/storage/redis.go
```diff
@@ -62,9 +68,56 @@ func (b *RedisDB) Close() error {
...
 }
 
+func (b *RedisDB) Execute(ctx context.Context, businessLogicFunc BusinessLogicFunc, watchKeys []string) (any, error) {
+	var finalOutput any
+	// Transactional function.
+	txf := func(tx *goredislib.Tx) error {
+		// Operation is commited only if the watched keys remain unchanged.
+		_, err := tx.TxPipelined(ctx, func(pipe goredislib.Pipeliner) error {
+
+			pipeCTX := context.WithValue(ctx, TX, pipe)
```
andresuribe87, line 78: Passing parameters around with context is discouraged.  How about the following, so you don't need to create this additional context?

As an additional benefit, it becomes very clear to the users of the `storage.ServiceStorage` what functions they can use inside the transaction.

```go
// in storage.go

func BusinessLogicFunc func(ctx context.Context, tx Tx) (any, error)

type ServiceStorage interface {
  // ...
  Execute(ctx context.Context, businessLogicFunc BusinessLogicFunc, watchKeys []string) (any, error)
}

type Tx interface {
  Read(ctx, namespace, key string) ([]byte, error)
  Write(ctx context.Context, namespace, key string) error
}

// in redis.go
type redisTx struct {
  pipe goredislib.Pipeliner
}

func (tx *redisTx) Read(ctx, namespace, key string, watchKeys []string) ([]byte, error) {
  return tx.pipe.Get(ctx, getRedisKey(namespace, key))
}

func (tx *redisTx) Write(ctx context.Context, namespace, key string) error {
  return tx.pipe.Set(ctx, getRedisKey(namespace, key)).Err()
}

func (b *RedisDB) Exec(ctx context.Context, bFun BusinessLogicFunc) (any, error) {
  txf := func(tx *goredislib.Tx) error {
    _, err := tx.TxPipelined(ctx, func(pipe goredislib.Pipeliner) error {
      var err error
      finalOutput, err = bFun(ctx, redisTx{pipe})
      return err
    })
    return err
  }
  b.db.Watch(ctx, txf, watchKeys...)
}

File: integration/credential_revocation_integration_test.go [hunk 12]
```diff
@@ -142,5 +142,6 @@ func TestRevocationCheckStatusListCredentialIntegration(t *testing.T) {
 	assert.NoError(t, err)
 	assert.NotEmpty(t, encodedListOriginal)
 
-	assert.NotEqual(t, encodedListOriginal.(string), encodedList)
+	// TODO: Fix Revocation Bug
+	// assert.NotEqual(t, encodedListOriginal.(string), encodedList)
```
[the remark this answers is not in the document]
nitro-neal: Added

pkg/service/credential/storage.go
```diff
@@ -163,6 +165,10 @@ func (cs *Storage) IncrementStatusListIndex(ctx context.Context) error {
 	return nil
 }
 
+func (cs *Storage) GetIncrementStatusListIndexWriteContextWatchKey() string {
+	return fmt.Sprintf("%v:%v", statusListIndexNamespace, currentListIndexKey)
```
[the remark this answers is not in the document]
nitro-neal: Yea this is fine, 
statusListIndexNamespace, currentListIndexKey

These variables will come from the application logic layer, so now its a bit the other way, application logic leaking into db.

I like this structure better though, I'll make it like this

pkg/storage/redis.go
```diff
@@ -62,9 +77,51 @@ func (b *RedisDB) Close() error {
...
+			}
+			return nil
+		})
+		return err
+	}
+
+	watchKeysStr := make([]string, 0)
+
+	for _, wc := range watchKeys {
+		watchKeysStr = append(watchKeysStr, fmt.Sprintf("%v:%v", wc.Namespace, wc.Key))
```
andresuribe87, line 101:
```suggestion
		watchKeysStr = append(watchKeysStr, getRedisKey(wc.Namespace, wc.Key))
```

pkg/storage/redis.go
```diff
@@ -62,9 +77,51 @@ func (b *RedisDB) Close() error {
...
+
+	expBackoff := backoff.NewExponentialBackOff()
+	expBackoff.MaxElapsedTime = MaxElapsedTime
+
+	err := backoff.Retry(func() error {
+		err := b.db.Watch(ctx, txf, watchKeysStr...)
+		if err != nil && errors.Is(err, goredislib.TxFailedErr) {
+			logrus.Warn("Optimistic lock lost. Retrying..")
+		}
+		return err
```
andresuribe87, line 112: Perhaps we shouldn't retry when the error is different than `TxFailedErr`. You can do so as follows:

```suggestion
		if err != nil && errors.Is(err, goredislib.TxFailedErr) {
			logrus.Warn("Optimistic lock lost. Retrying..")
			return err
		}
		return backoff.Permanent(err)
```

andresuribe87 (approved):
Big big fan of how this turned out. This is looking awesome, thanks!

A few minor suggestions.

pkg/storage/redis.go
nitro-neal: ok that makes sense,

pkg/service/credential/storage.go
```diff
@@ -150,6 +148,19 @@ func (cs *Storage) WriteMany(ctx context.Context, writeContexts []WriteContext)
 	return cs.db.WriteMany(ctx, namespaces, keys, values)
 }
 
+func (cs *Storage) IncrementStatusListIndexTx(ctx context.Context, tx storage.Tx) error {
```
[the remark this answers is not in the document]
nitro-neal: 🤔  All the other methods that are called form the storage files are Exported, they definitely don't need to be, but for matching the rest of the code I'll keep them Uppercase for now (since they are called in the same way as every other method in the storage files).

We can go through and change all to not exported in future PR if needed

PR #273, opened 2023-02-01
Enable cors as an option. Docker default to allowing all.

andresuribe87:
With this change, you can request resources from a browser after running `mage run`

### DIFF
--- a/integration/common.go
+++ b/integration/common.go
@@ -353,16 +353,9 @@ func getJSONElement(jsonString string, jsonPath string) (string, error) {
 func get(url string) (string, error) {
 	logrus.Println(fmt.Sprintf("\nPerforming GET request to:  %s\n", url))
 
-	req, err := http.NewRequest(http.MethodGet, url, nil)
+	resp, err := http.Get(url) // #nosec: testing only.
 	if err != nil {
-		return "", errors.Wrap(err, "building http req")
-	}
-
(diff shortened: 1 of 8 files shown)


integration/common.go
```diff
@@ -353,9 +353,16 @@ func getJSONElement(jsonString string, jsonPath string) (string, error) {
...
 	logrus.Println(fmt.Sprintf("\nPerforming GET request to:  %s\n", url))
 
-	resp, err := http.Get(url) // #nosec: testing only.
+	req, err := http.NewRequest(http.MethodGet, url, nil)
 	if err != nil {
-		return "", errors.Wrapf(err, "getting url: %s", url)
+		return "", errors.Wrap(err, "building http req")
+	}
+
+	req.Header.Set("Origin", "http://foo.com")
```
nitro-neal, line 361: did you mean to do this for every get() request?

nitro-neal (approved):
Awesome! Did you try to do a req from a react app or something for the 'real deal' test?

decentralgabe (approved):

integration/common.go
andresuribe87: Actually removed this. An actual test would need to check the presence of a header.

andresuribe87:
> Awesome! Did you try to do a req from a react app or something for the 'real deal' test?

I did a request from the chrome dev tools in `localhost:12345`

PR #274, opened 2023-02-02
adding exists to db interface and other error handling

nitro-neal:
# Overview
* Handle case where `current-list-index >= bitStringLength`
* Only generate the new statusListIndex, and array after doing a read and making sure it does not exist


.github/workflows/steelthread.yml
```diff
@@ -11,7 +11,7 @@ on:
 
 jobs:
   end2endtest:
-    timeout-minutes: 15
+    timeout-minutes: 20
```
andresuribe87, line 14: Under what scenarios does this take more than 15 min?

.github/workflows/steelthread.yml
```diff
@@ -30,7 +30,12 @@ jobs:
       run: docker-compose -f "build/docker-compose.yml" up -d --build
 
     - name: End To End Test
-      run: mage integration
+      uses: nick-fields/retry@v2
+      with:
+        timeout_minutes: 15
+        max_attempts: 3
+        retry_on: error
+        command: mage integration
```
andresuribe87, line 38: I'm concerned that doing retrys would hide flakiness and real issues we should address. Thoughts?

pkg/storage/db_test.go
```diff
@@ -199,6 +199,37 @@ func TestDBEmptyNamespace(t *testing.T) {
...
+		err = db.Write(context.Background(), namespace, key, nil)
+		assert.NoError(t, err)
+
+		dummyData := []byte("dummy")
+		err = db.Write(context.Background(), namespace, key, dummyData)
+		assert.NoError(t, err)
+
+		exists, err = db.Exists(context.Background(), namespace, key)
+		assert.NoError(t, err)
+		assert.True(t, exists)
```
andresuribe87, line 229: Would recommend splitting these into three tests.

pkg/service/credential/storage.go
```diff
@@ -80,26 +80,42 @@ type StatusListIndex struct {
...
+	if err != nil {
+		return nil, util.LoggingErrorMsg(err, "exists for statusListIndexesKey")
 	}
 
-	statusListIndexBytes, err := json.Marshal(StatusListIndex{Index: 0})
-	if err != nil {
-		return nil, util.LoggingErrorMsg(err, "could not marshal status list index bytes")
+	if listIndexKeyExists != statusListIndexesKeyExists {
+		return nil, util.LoggingNewError("list index and status list index not in the same state")
 	}
```
andresuribe87, line 99: What behavior is this changing? Would recommend adding a test.

.github/workflows/steelthread.yml
nitro-neal: should never happen, it needs the timeout-minutes field to use the action module

.github/workflows/steelthread.yml
nitro-neal: It does fail like 10% of the time, I think it has to do with the nondetermanistic threads test. I have retry logic on our client, but maybe there is some better handling we can do so the failure goes down to like 1% of the time

pkg/storage/db_test.go
nitro-neal: ok, lots more lines of code though! 🤣

.github/workflows/steelthread.yml
nitro-neal: removed this, yea lets keep it vanilla

pkg/server/router/credential_test.go
```diff
@@ -275,4 +275,72 @@ func TestCredentialRouter(t *testing.T) {
 		assert.NotEmpty(tt, credStatusMapTwo["statusListIndex"])
 
 	})
+
+	t.Run("Credential Status List Test On Create", func(tt *testing.T) {
```
andresuribe87, line 279: It's unclear what **behavior** this test is exercising. Would recommend following this [naming best practices](https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices#naming-your-tests) for this test as well. Perhaps worth also splitting it up.

pkg/storage/db_test.go
```diff
@@ -199,6 +199,72 @@ func TestDBEmptyNamespace(t *testing.T) {
...
 
+func TestDBExistsNoNamespaceKey(t *testing.T) {
+	for _, dbImpl := range getDBImplementations(t) {
+		db := dbImpl
+
+		exists, err := db.Exists(context.Background(), "dnenamespace", "dnekey")
+		assert.NoError(t, err)
+		assert.False(t, exists)
+	}
+}
```
andresuribe87, line 210: This is awesome! Small suggestion to also include the expected behavior in the test, so reader can quickly verify that what you're asserting it correct.
```go
func TestDBExists_FalseWhenNoNamespaceKey(t *testing.T) {
	for _, dbImpl := range getDBImplementations(t) {
		db := dbImpl

		exists, err := db.Exists(context.Background(), "dnenamespace", "dnekey")
		assert.NoError(t, err)
		assert.False(t, exists)
	}
}
```

pkg/storage/db_test.go
```diff
@@ -199,6 +199,72 @@ func TestDBEmptyNamespace(t *testing.T) {
...
+
+func TestDBExistsNoKey(t *testing.T) {
```
andresuribe87, line 212:
```suggestion
func TestDBExists_FalseWhenKeyIsAbsent(t *testing.T) {
```

pkg/storage/db_test.go
```diff
@@ -199,6 +199,72 @@ func TestDBEmptyNamespace(t *testing.T) {
...
+		assert.NoError(t, err)
+
+		exists, err := db.Exists(context.Background(), namespace, "otherdnekey")
+		assert.NoError(t, err)
+		assert.False(t, exists)
+
+	}
+}
+
+func TestDBExistsNilValue(t *testing.T) {
```
andresuribe87, line 228: You get the point :)

pkg/storage/db_test.go
```diff
@@ -199,6 +199,72 @@ func TestDBEmptyNamespace(t *testing.T) {
...
+		assert.True(t, exists)
+
+		err = db.Delete(context.Background(), namespace, key)
+		assert.NoError(t, err)
+
+		exists, err = db.Exists(context.Background(), namespace, key)
+		assert.NoError(t, err)
+		assert.False(t, exists)
+	}
+}
```
andresuribe87, line 266: Would recommend naming `TestDBExists_FalseAfterDeletingKey`

pkg/service/credential/storage.go
```diff
@@ -80,26 +80,42 @@ type StatusListIndex struct {
...
 
-	if err := db.Write(context.Background(), statusListIndexNamespace, currentListIndexKey, statusListIndexBytes); err != nil {
-		return nil, util.LoggingErrorMsg(err, "problem writing current list index to db")
+	if !listIndexKeyExists && !statusListIndexesKeyExists {
```
decentralgabe, line 100: nit: consider a utility variable to help make this more readable

like `indexKeysDontExist := !listIndexKeyExists && !statusListIndexesKeyExists`

decentralgabe (approved):

PR #275, opened 2023-02-03
Fix docs build

andresuribe87:
The publishing of swagger docs is broken in master. This fixes it. Additionally a lot of automated formatting going on.


nitro-neal (approved):
good stuff

PR #280, opened 2023-02-07
Credential suspension

nitro-neal:
# Overview
This change allows the ssi-service to have suspension credentials

# Description
Following the status list spec https://w3c.github.io/vc-status-list-2021/ we now have support for revocable and suspendable credentials. This change enforces and assumes that any one credential can only be revocable OR suspendable and not both. This fits the spec as in the example there is only room for one credentialStatus object and not an array

```
{
  "@context": [
    "https://www.w3.org/2018/credentials/v1",
    "https://w3id.org/vc/status-list/2021/v1"
  ],
  "id": "https://example.com/credentials/23894672394",
  "type": ["VerifiableCredential"],
  "issuer": "did:example:12345",
  "issued": "2021-04-05T14:27:42Z",
  "credentialStatus": {
    "id": "https://example.com/credentials/status/3#94567"
    "type": "StatusList2021Entry",
    "statusPurpose": "revocation",
    "statusListIndex": "94567",
    "statusListCredential": "https://example.com/credentials/status/3"
  },
  "credentialSubject": {
    "id": "did:example:6789",
    "type": "Person"
  },
  "proof": { ... }
}
```

 
The StatusListCredential are now unique to 3 fields <issuer,schema,statusPurpose>
issuer1,schema1,revocation = StatusList2021Credential1
issuer1,schema1,suspension = StatusList2021Credential2

### DIFF
--- a/pkg/server/router/credential_test.go
+++ b/pkg/server/router/credential_test.go
@@ -602,7 +602,7 @@ func TestCredentialRouter(t *testing.T) {
 		assert.Equal(tt, statusListIndexes, statusListIndexesAfterRestart)
 	})
 
-	t.Run("Test Create Suspendable Credential", func(tt *testing.T) {
+	t.Run("Create Suspendable Credential", func(tt *testing.T) {
 		issuer, schema, credService := createCredServicePrereqs(tt)
 		subject := "did:test:345"
 
@@ -656,7 +656,7 @@ func TestCredentialRouter(t *testing.T) {
 		assert.False(tt, valid)
 	})
 
-	t.Run("Test Update Suspendable Credential", func(tt *testing.T) {
+	t.Run("Update Suspendable Credential To Suspended", func(tt *testing.T) {
 		issuer, schema, credService := createCredServicePrereqs(tt)
 		subject := "did:test:345"
 
--- a/pkg/server/router/credential_test.go
+++ b/pkg/server/router/credential_test.go
@@ -601,4 +601,244 @@ func TestCredentialRouter(t *testing.T) {
 
 		assert.Equal(tt, statusListIndexes, statusListIndexesAfterRestart)
 	})
+
+	t.Run("Test Create Suspendable Credential", func(tt *testing.T) {
+		issuer, schema, credService := createCredServicePrereqs(tt)
+		subject := "did:test:345"
+
+		createdCred, err := credService.CreateCredential(context.Background(), credential.CreateCredentialRequest{
+			Issuer:     issuer,
+			Subject:    subject,
+			JSONSchema: schema,
+			Data: map[string]any{
+				"email": "Satoshi@Nakamoto.btc",
+			},
+			Expiry:      time.Now().Add(24 * time.Hour).Format(time.RFC3339),
+			Suspendable: true,
+		})
+
+		assert.NoError(tt, err)
+		assert.NotEmpty(tt, createdCred)
+		assert.NotEmpty(tt, createdCred.CredentialJWT)
+
+		statusBytes, err := json.Marshal(createdCred.Credential.CredentialStatus)
+		assert.NoError(tt, err)
+
+		var statusEntry status.StatusList2021Entry
+		err = json.Unmarshal(statusBytes, &statusEntry)
+		assert.NoError(tt, err)
+
+		assert.Contains(tt, statusEntry.ID, fmt.Sprintf("http://localhost:1234/v1/credentials/%s/status", createdCred.ID))
+		assert.Contains(tt, statusEntry.StatusListCredential, "http://localhost:1234/v1/credentials/status")
+		assert.NotEmpty(tt, statusEntry.StatusListIndex)
+
+		credStatus, err := credService.GetCredentialStatus(context.Background(), credential.GetCredentialStatusRequest{ID: createdCred.ID})
+		assert.NoError(tt, err)
+		assert.Equal(tt, credStatus.Revoked, false)
+		assert.Equal(tt, credStatus.Suspended, false)
+
+		credStatusListStr := statusEntry.StatusListCredential
+
+		credStatusListID := strings.Split(credStatusListStr, "/")[len(strings.Split(credStatusListStr, "/"))-1]
+		credStatusList, err := credService.GetCredentialStatusList(context.Background(), credential.GetCredentialStatusListRequest{ID: credStatusListID})
+		assert.NoError(tt, err)
+		assert.Equal(tt, credStatusList.Credential.ID, statusEntry.StatusListCredential)
+
+		credentialSubject := credStatusList.Container.Credential.CredentialSubject
+		assert.NotEmpty(tt, credentialSubject)
+
+		encodedList := credentialSubject["encodedList"]
+		assert.NotEmpty(tt, encodedList)
+
+		// Validate the StatusListIndex in not flipped in the credStatusList
+		valid, err := status.ValidateCredentialInStatusList(*createdCred.Credential, *credStatusList.Credential)
+		assert.NoError(tt, err)
+		assert.False(tt, valid)
+	})
+
+	t.Run("Test Update Suspendable Credential", func(tt *testing.T) {
+		issuer, schema, credService := createCredServicePrereqs(tt)
+		subject := "did:test:345"
+
+		createdCred, err := credService.CreateCredential(context.Background(), credential.CreateCredentialRequest{
+			Issuer:     issuer,
+			Subject:    subject,
+			JSONSchema: schema,
+			Data: map[string]any{
+				"email": "Satoshi@Nakamoto.btc",
+			},
+			Expiry:      time.Now().Add(24 * time.Hour).Format(time.RFC3339),
+			Suspendable: true,
+		})
+
+		assert.NoError(tt, err)
+		assert.NotEmpty(tt, createdCred)
+		assert.NotEmpty(tt, createdCred.CredentialJWT)
+
+		statusBytes, err := json.Marshal(createdCred.Credential.CredentialStatus)
+		assert.NoError(tt, err)
+
+		var statusEntry status.StatusList2021Entry
+		err = json.Unmarshal(statusBytes, &statusEntry)
+		assert.NoError(tt, err)
+
+		assert.Contains(tt, statusEntry.ID, fmt.Sprintf("http://localhost:1234/v1/credentials/%s/status", createdCred.ID))
+		assert.Contains(tt, statusEntry.StatusListCredential, "http://localhost:1234/v1/credentials/status")
+		assert.NotEmpty(tt, statusEntry.StatusListIndex)
+
+		credStatus, err := credService.GetCredentialStatus(context.Background(), credential.GetCredentialStatusRequest{ID: createdCred.ID})
+		assert.NoError(tt, err)
+		assert.Equal(tt, credStatus.Revoked, false)
+		assert.Equal(tt, credStatus.Suspended, false)
+
+		credStatusListStr := statusEntry.StatusListCredential
+
+		credStatusListID := strings.Split(credStatusListStr, "/")[len(strings.Split(credStatusListStr, "/"))-1]
+		credStatusList, err := credService.GetCredentialStatusList(context.Background(), credential.GetCredentialStatusListRequest{ID: credStatusListID})
+		assert.NoError(tt, err)
+		assert.Equal(tt, credStatusList.Credential.ID, statusEntry.StatusListCredential)
+
+		credentialSubject := credStatusList.Container.Credential.CredentialSubject
+		assert.NotEmpty(tt, credentialSubject)
+
+		encodedList := credentialSubject["encodedList"]
+		assert.NotEmpty(tt, encodedList)
+
+		// Validate the StatusListIndex in not flipped in the credStatusList
+		valid, err := status.ValidateCredentialInStatusList(*createdCred.Credential, *credStatusList.Credential)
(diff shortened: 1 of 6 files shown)


andresuribe87:
> This change enforces and assumes that any one credential can only be revocable OR suspendable and not both

I actually had this question recently, and it's in the process of being clarified in the spec. See https://github.com/w3c/vc-status-list-2021/issues/37#issuecomment-1416348418 for details. Turns out that `credentialStatus` can be an object, or an array.

pkg/server/router/credential_test.go
```diff
@@ -601,4 +601,244 @@ func TestCredentialRouter(t *testing.T) {
...
+		assert.Nil(tt, updatedStatus)
+		assert.Error(tt, err)
+		assert.ErrorContains(tt, err, "has a different status purpose<revocation> value than the status credential<suspension>")
+	})
+
+}
+
+func createCredServicePrereqs(tt *testing.T) (string, string, credential.Service) {
+	bolt := setupTestDB(tt)
+	assert.NotNil(tt, bolt)
```
decentralgabe, line 808: nit: you may want to change these asserts to `require` in setup so that the test fails fast if the prereqs aren't there

pkg/service/credential/model.go
```diff
@@ -60,16 +61,19 @@ type GetCredentialStatusRequest struct {
 }
 
 type GetCredentialStatusResponse struct {
```
decentralgabe, line 63: is there a way to enforce mutual exclusion on these statuses? perhaps a validate method on each of them?

pkg/service/credential/service.go
```diff
@@ -116,6 +116,10 @@ func (s Service) createCredentialFunc(request CreateCredentialRequest) storage.B
 func (s Service) createCredentialBusinessLogic(ctx context.Context, request CreateCredentialRequest, tx storage.Tx) (*CreateCredentialResponse, error) {
 	logrus.Debugf("creating credential: %+v", request)
 
+	if request.Revocable && request.Suspendable {
+		return nil, util.LoggingNewError("credential cannot be revocable and suspendable")
```
decentralgabe, line 120: maybe: "credential may have at most one status"

pkg/service/credential/service.go
```diff
@@ -567,14 +584,22 @@ func updateCredentialStatus(ctx context.Context, s Service, gotCred *StoredCrede
 		return nil, util.LoggingNewErrorf("problem with getting status list credential for issuer: %s schema: %s", gotCred.Issuer, gotCred.Schema)
 	}
 
-	var revokedStatusCreds []credential.VerifiableCredential
+	var revokedOrSuspendedStatusCreds []credential.VerifiableCredential
```
decentralgabe, line 587: nit: `statusCreds`

pkg/service/credential/storage.go
```diff
@@ -223,16 +227,55 @@ func (cs *Storage) StoreCredential(ctx context.Context, request StoreCredentialR
 	return cs.storeCredential(ctx, request, credentialNamespace)
 }
 
-func (cs *Storage) StoreStatusListCredentialTx(ctx context.Context, request StoreCredentialRequest, tx storage.Tx) error {
-	wc, err := cs.getStoreCredentialWriteContext(request, statusListCredentialNamespace)
+func (cs *Storage) StoreStatusListCredentialTx(ctx context.Context, request StoreCredentialRequest, statusPurpose statussdk.StatusPurpose, tx storage.Tx) error {
```
decentralgabe, line 230: nit: may make sense for tx storage.Tx to come as the 2nd parameter here and elsewhere

pkg/service/credential/storage.go
```diff
@@ -489,8 +525,46 @@ func (cs *Storage) GetCredentialsByIssuerAndSchema(ctx context.Context, issuer s
...
+		logrus.Warnf("no credentials found for issuer: %s and schema %s", util.SanitizeLog(issuer), util.SanitizeLog(schema))
+		return nil, nil
+	}
+
+	// now get each credential by key
+	var storedCreds []StoredCredential
+	for _, key := range issuerSchemaKeys {
+		credBytes, err := cs.db.Read(ctx, statusListCredentialNamespace, key)
+		if err != nil {
+			logrus.WithError(err).Errorf("could not read credential with key: %s", key)
```
decentralgabe, line 552: return err here?

decentralgabe (commented):
nice!

pkg/server/router/credential_test.go
```diff
@@ -601,4 +601,244 @@ func TestCredentialRouter(t *testing.T) {
...
+		assert.NotEmpty(tt, statusEntry.StatusListIndex)
+
+		credStatus, err := credService.GetCredentialStatus(context.Background(), credential.GetCredentialStatusRequest{ID: createdCred.ID})
+		assert.NoError(tt, err)
+		assert.Equal(tt, credStatus.Revoked, false)
+		assert.Equal(tt, credStatus.Suspended, false)
+
+		credStatusListStr := statusEntry.StatusListCredential
+
+		credStatusListID := strings.Split(credStatusListStr, "/")[len(strings.Split(credStatusListStr, "/"))-1]
```
andresuribe87, line 642: Might be easier to grok
```suggestion
		_, credStatusListID, ok := strings.Cut(credStatusListStr, "/v1/credentials/status/")
		assert.True(tt, ok)
```

pkg/server/router/credential_test.go
```diff
@@ -601,4 +601,244 @@ func TestCredentialRouter(t *testing.T) {
...
+		assert.Equal(tt, credStatusList.Credential.ID, statusEntry.StatusListCredential)
+
+		credentialSubject := credStatusList.Container.Credential.CredentialSubject
+		assert.NotEmpty(tt, credentialSubject)
+
+		encodedList := credentialSubject["encodedList"]
+		assert.NotEmpty(tt, encodedList)
+
+		// Validate the StatusListIndex in not flipped in the credStatusList
+		valid, err := status.ValidateCredentialInStatusList(*createdCred.Credential, *credStatusList.Credential)
```
andresuribe87, line 654: I still find this function name so confusing :/

pkg/server/router/credential_test.go
```diff
@@ -601,4 +601,244 @@ func TestCredentialRouter(t *testing.T) {
...
+		err = json.Unmarshal(statusBytes, &statusEntry)
+		assert.NoError(tt, err)
+
+		assert.Contains(tt, statusEntry.ID, fmt.Sprintf("http://localhost:1234/v1/credentials/%s/status", createdCred.ID))
+		assert.Contains(tt, statusEntry.StatusListCredential, "http://localhost:1234/v1/credentials/status")
+		assert.NotEmpty(tt, statusEntry.StatusListIndex)
+
+		credStatus, err := credService.GetCredentialStatus(context.Background(), credential.GetCredentialStatusRequest{ID: createdCred.ID})
+		assert.NoError(tt, err)
+		assert.Equal(tt, credStatus.Revoked, false)
```
andresuribe87, line 691: Since you aren't testing things related to `Revoked`, consider removing all `Revoked` related code in this test. 
```suggestion
```

pkg/server/router/credential_test.go
```diff
@@ -601,4 +601,244 @@ func TestCredentialRouter(t *testing.T) {
...
+		assert.Equal(tt, credStatus.Suspended, false)
+
+		credStatusListStr := statusEntry.StatusListCredential
+
+		credStatusListID := strings.Split(credStatusListStr, "/")[len(strings.Split(credStatusListStr, "/"))-1]
```
andresuribe87, line 696: Same suggestion as above :)

pkg/server/router/credential_test.go
```diff
@@ -601,4 +601,244 @@ func TestCredentialRouter(t *testing.T) {
...
+		assert.NoError(tt, err)
+		assert.Equal(tt, credStatusList.Credential.ID, statusEntry.StatusListCredential)
+
+		credentialSubject := credStatusList.Container.Credential.CredentialSubject
+		assert.NotEmpty(tt, credentialSubject)
+
+		encodedList := credentialSubject["encodedList"]
+		assert.NotEmpty(tt, encodedList)
+
+		// Validate the StatusListIndex in not flipped in the credStatusList
```
andresuribe87, line 707:
```suggestion
		// Validate the StatusListIndex is not flipped in the credStatusList
```

pkg/service/credential/storage.go
```diff
@@ -223,16 +227,55 @@ func (cs *Storage) StoreCredential(ctx context.Context, request StoreCredentialR
...
+func (cs *Storage) StoreStatusListCredentialTx(ctx context.Context, request StoreCredentialRequest, statusPurpose statussdk.StatusPurpose, tx storage.Tx) error {
+	if !request.IsValid() {
+		return util.LoggingNewError("store request request is not valid")
+	}
+
+	// transform the credential into its denormalized form for storage
+	storedCredential, err := buildStoredCredential(request)
 	if err != nil {
-		return errors.Wrap(err, "could not get stored credential write context")
+		return errors.Wrap(err, "could not build stored credential")
```
andresuribe87, line 238: While you're here :) 
```suggestion
		return errors.Wrap(err, "building stored credential")
```

pkg/service/credential/storage.go
```diff
@@ -223,16 +227,55 @@ func (cs *Storage) StoreCredential(ctx context.Context, request StoreCredentialR
...
+		credBytes = v
+		break
+	}
+	if len(credBytes) == 0 {
+		return nil, util.LoggingNewErrorf("could not get credential from storage %s with id: %s", credentialNotFoundErrMsg, id)
+	}
+
+	var stored StoredCredential
+	if err = json.Unmarshal(credBytes, &stored); err != nil {
+		return nil, util.LoggingErrorMsgf(err, "could not unmarshal stored credential: %s", id)
```
andresuribe87, line 276:
```suggestion
		return nil, util.LoggingErrorMsgf(err, "unmarshalling stored credential: %s", id)
```

pkg/service/credential/storage.go
```diff
@@ -489,8 +525,46 @@ func (cs *Storage) GetCredentialsByIssuerAndSchema(ctx context.Context, issuer s
...
+				logrus.WithError(err).Errorf("could not unmarshal credential with key: %s", key)
+			}
+
+			storedCreds = append(storedCreds, cred)
+		}
+	}
+
+	if len(storedCreds) == 0 {
+		logrus.Warnf("no credentials able to be retrieved for issuer: %s", issuerSchemaKeys)
+	}
```
andresuribe87, line 565: What's the advantage of warning here, vs doing nothing?

pkg/service/credential/model.go
nitro-neal: Yea that is better:

Added:
```
func (csr CreateCredentialRequest) isStatusValid() bool {
	if csr.Revocable && csr.Suspendable {
		return false
	}
	return true
}
```

decentralgabe: `return !(csr.Revocable && csr.Suspendable)`

pkg/service/credential/storage.go
nitro-neal: Hmm this is the the same for a lot of the original functions we have in this same file:
GetCredentialsBySchema, GetCredentialsBySubject etc..

The current functionality is if there are 100 creds and one fails to read it will return 99, 

probably we want this to return all 100 or nothing right? I can make changes to all areas where it is like this

decentralgabe: ah let's keep as is

pkg/server/router/credential_test.go
```diff
@@ -601,4 +601,244 @@ func TestCredentialRouter(t *testing.T) {
...
+		assert.NotEmpty(tt, createdCred.CredentialJWT)
+
+		statusBytes, err := json.Marshal(createdCred.Credential.CredentialStatus)
+		assert.NoError(tt, err)
+
+		var statusEntry status.StatusList2021Entry
+		err = json.Unmarshal(statusBytes, &statusEntry)
+		assert.NoError(tt, err)
+
+		assert.Contains(tt, statusEntry.ID, fmt.Sprintf("http://localhost:1234/v1/credentials/%s/status", createdCred.ID))
```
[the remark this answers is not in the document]
nitro-neal: In the spec the statusEntry.ID  needs to be a link to the status basically
would look like this after you revoke it:

GET
```
http://localhost:8080/v1/credentials/f91c2bc9-1688-471d-bca1-156ee18514cd/status
```

```
{"revoked":true}
```

pkg/service/credential/storage.go
nitro-neal: As posted above: 

This is the the same for a lot of the original functions we have in this same file:
GetCredentialsBySchema, GetCredentialsBySubject etc..

The current functionality is if there are 100 creds and one fails to read it will return 99,

probably we want this to return all 100 or nothing right? I can make changes to all areas where it is like this.

nitro-neal: I'm inclined to make the changes, may break a few unit tests and change functionality

nitro-neal: Not too much, just for debugging we can see that people are querying a cred for a certain issuer that does not exist i suppose

pkg/server/router/credential_test.go
decentralgabe: `ValidateCredentialStatus`

pkg/service/credential/storage.go
decentralgabe: think it's useful to keep

PR #293, opened 2023-02-20
Revoke Key API 

andorsk:
# Overview
Revoke Key APIs

# Description
Revoke a key API

# How Has This Been Tested?
- [ ] Simple Test [Here](pkg/server/router/keystore_test.go) There is the possibility to add more tests, but with 24% code overage across the entire repo, I didn't spend a ton of time worrying about anything other than the basic test vector.

# Checklist


pkg/server/router/keystore.go
```diff
@@ -131,3 +131,33 @@ func (ksr *KeyStoreRouter) GetKeyDetails(ctx context.Context, w http.ResponseWri
...
+	}
+
+	_, err := ksr.service.DeleteKey(ctx, keystore.DeleteKeyRequest{ID: *id})
+	if err != nil {
+		errMsg := fmt.Sprintf("could not delete key for id: %s", *id)
+		logrus.WithError(err).Error(errMsg)
+		return framework.NewRequestError(errors.Wrap(err, errMsg), http.StatusBadRequest)
+	}
+
+	resp := GetKeyDetailsResponse{}
```
decentralgabe, line 161:
```suggestion
	var resp GetKeyDetailsResponse
```

pkg/service/keystore/model.go
```diff
@@ -35,3 +35,10 @@ type GetKeyDetailsResponse struct {
 	Controller string
 	CreatedAt  string
 }
+
+type DeleteKeyRequest struct {
+	ID string
+}
+
+type DeleteKeyResponse struct {
```
decentralgabe, line 43: is this necessary?

pkg/service/keystore/service.go
```diff
@@ -126,6 +126,17 @@ func (s Service) GetKey(ctx context.Context, request GetKeyRequest) (*GetKeyResp
 	}, nil
 }
 
+func (s Service) DeleteKey(ctx context.Context, request DeleteKeyRequest) (*DeleteKeyResponse, error) {
+	logrus.Debugf("deleteing key: %+v", request)
+	id := request.ID
+	err := s.storage.DeleteKey(ctx, id)
+	if err != nil {
```
decentralgabe, line 133:
```suggestion
	if err := s.storage.DeleteKey(ctx, id); err != nil {
```

pkg/service/keystore/service.go
```diff
@@ -126,6 +126,17 @@ func (s Service) GetKey(ctx context.Context, request GetKeyRequest) (*GetKeyResp
 	}, nil
 }
 
+func (s Service) DeleteKey(ctx context.Context, request DeleteKeyRequest) (*DeleteKeyResponse, error) {
```
decentralgabe, line 129: since there's no response I think this should just return `error`

decentralgabe (changes requested):

andresuribe87:
@andorsk would you mind describing the use case you're addressing with adding Deletion?

andorsk:
@andresuribe87  sure: At a high level, just offering the option to delete a key was something I was using while testing things. Make key, delete it after. I also just don't see a great reason to _not_ support this, as it's a basic CRUD op with keys that I'd expect should be supportable unless there's not a good reason not to?

pkg/service/keystore/service.go
andorsk: Sure thing. I think I was trying to be consistent with some other stuff I've seen somewhere in the ssi-server, but throwing just an error is fine too .

decentralgabe:
Soft delete would be *safer* - meaning set a deleted flag in the DB but retain the key just in case. you can get into some hairy situations if you accidentally delete a key

andorsk:
@decentralgabe agreed there are different ways to implement this. I was going for a simple implementation. We could easily add a field saying "Hide" into the `StoredKey` object if we want to. For my use case, actually deleting the key was sufficient and what I wanted to do.

andorsk:
@decentralgabe changed from delete key to revoke key.

pkg/service/keystore/model.go
andorsk: I've updated the following to just send an error back instead of a response.

andorsk:
@decentralgabe change implementation to add "revocation" field.  Is e2e test passing required? Otherwise, would like feedback here if possible.

decentralgabe:
@andorsk yes getting the e2e test is a requirement. It seems to be passing on the main branch and other PRs so it must be with something you introduced. If you can prove otherwise let me know 😄

decentralgabe:
@andorsk would you mind adding a `revokedAt` field?

andorsk:
thanks @decentralgabe . Yep. Adding a revoked at field is a good callout. I'll add it. 

Regarding e2e, I'm wondering if there's possible some dependency that can break in the chain for e2e testing suite ( maybe why it's not "required" on the test suite? I.e e2e started breaking at [a576bd9](https://github.com/TBD54566975/ssi-service/pull/293/commits/a576bd958d004b1a9302cc48adf6d76798252609), which was just a merge into the main branch, so it's strange to me e2e started breaking. I can look into this a little more.

andorsk:
lol. CI/CD testing just completed it started passing again, after adding the `revokedAt` field. I have a hard time believing adding a revoked field causes a failing test to pass, but I could be wrong here. It sounds like to me the `e2e` testing had some non-code dependencies that maybe were unavailable yesterday but maybe someone knows better?

decentralgabe (approved):
lgtm - thanks for the changes.

PR #294, opened 2023-02-21
added utility on new key store service from keystorage

andorsk:
# Overview
Helper to build a keystorage service form an existing keystore.

# Description
I wanted to use an existing keystore.keyStorage for `ssi-service`. The main issue is that https://github.com/TBD54566975/ssi-service/blob/main/pkg/service/keystore/service.go#L51 generates a new service key, and I wanted to read an existing service key stored in a db. It wasn't clear how to spin up a KeystoreService with an existing ServiceKey, and the storage and config methods are privately scoped, so other applications can't read it. Therefore, I decided to add a helper method that adds a way to create a new keystore service with an existing storage configuration. 

# How Has This Been Tested?
I've confirmed this function works, and tested locally, but haven't built any tests for CI/CD. We are hanging ~24% test coverage right now, so not sure if a test is required, but can add one it if that is a requirement for merge. 

# Checklist

### DIFF
--- a/pkg/service/keystore/service.go
+++ b/pkg/service/keystore/service.go
@@ -46,6 +46,18 @@ func (s Service) Config() config.KeyStoreServiceConfig {
 	return s.config
 }
 
+// NewKeyStoreFromKeystoreStorage uses  a keystore service directly from storage object
+func NewKeyStoreFromKeystoreStorage(config config.KeyStoreServiceConfig, keyStoreStorage *Storage) (*Service, error) {
+	service := Service{
+		storage: keyStoreStorage,
+		config:  config,
+	}
+	if !service.Status().IsReady() {
+		return nil, errors.New(service.Status().Message)
+	}
+	return &service, nil
+}
+
 func NewKeyStoreService(config config.KeyStoreServiceConfig, s storage.ServiceStorage) (*Service, error) {
 	// First, generate a service key
 	serviceKey, serviceKeySalt, err := GenerateServiceKey(config.ServiceKeyPassword)


pkg/service/keystore/service.go
```diff
@@ -46,6 +46,22 @@ func (s Service) Config() config.KeyStoreServiceConfig {
 	return s.config
 }
 
+// NewKeyStoreFromKeystoreStorage uses  a keystore service directly from storage object
+func NewKeyStoreFromKeystoreStorage(config config.KeyStoreServiceConfig, keyStoreStorage *Storage) (*Service, error) {
```
decentralgabe, line 50: where would this key store be coming from in practice?

andorsk: @decentralgabe so it's a good question. I was writing some personal stuff for testing around this and I needed an entry point that did not generate a new service key and salt, which is why I needed a method to support this action. I'm happy to just support this entrypoint on my own fork of this, or if there's a better way to handle this I'm open to it.

decentralgabe: what about adding a parameter to the config to not generate a new key and instead look for an existing one?

andorsk: @decentralgabe I think that's separate concerns. How it's managed in the config (optional/default behavior) is certainly important, but in the end of the day you'd still need to provide the logic in the NewKeyStoreService function to ignore re-generating the service keys, and/or to load the existing one. You then need to address the behavior of how they are loaded internally, etc, which was out of scope for this PR.  Certainly a valid point, just not sure if that's addressed best somewhere else.

andorsk: I checked if `ssi-service` key was set, and if it was, loaded that, using the method above. There are other possible ways to handle this, which is why I didn't want to tackle the logic around configuration/loading existing keys in this PR.

decentralgabe:
Would like to get @andresuribe87 's perspective before continuing

pkg/service/keystore/service.go
andresuribe87: This parameters of this new constructor makes sense to me. I recommend only having a single constructor, instead of 2. It seems to me that it makes sense to update all callers to use the new constructor, and remove the old one.

Keeping a single constructor can then be named `NewKeyStoreService`.

andorsk: Yea..@andresuribe87 that works and I think is a reasonable way forward.

decentralgabe:
@andorsk are you still interested in pushing this along? if so what's needed?

andorsk:
@decentralgabe thanks. This is probably still a good idea, but I'm OK with backlogging it for now. Currently it's not a priority with me.

decentralgabe:
Thanks @andorsk I'll keep it open - agree it's useful.

PR #298, opened 2023-02-28
Adding db impl for webhooks

nitro-neal:
# Overview
This change adds the db implementation for webhooks

# Description
Webhooks are defined by having a noun verb pairing instead of id. This way one can easily see all webhooks accosiated with a noun verb pairing at any time. Also, since these noun-verb look ups will happen on every request it makes it a constant lookup on each request instead of scanning

# Routes
PUT 
http://localhost:8080/v1/webhooks
```
{
  "noun": "Manifest",
  "verb": "Create",
  "url": "https://www.tbd.website/"
}
```

Response
```
{
    "webhook": {
        "noun": "Manifest",
        "verb": "Create",
        "urls": [
            "https://www.tbd.website/"
        ]
    }
}
```

GET 
http://localhost:8080/v1/webhooks/Manifest/Create
```
{
    "webhook": {
        "noun": "Manifest",
        "verb": "Create",
        "urls": [
            "https://www.tbd.website/"
        ]
    }
}
```
Note that if you create another webhook with the same noun verb pairing and a different URL, and then do a GET you will get this type of response object. (One webhook object with many urls in array)

GET 
http://localhost:8080/v1/webhooks/Manifest/Create
```
{
    "webhook": {
        "noun": "Manifest",
        "verb": "Create",
        "urls": [
            "https://www.tbd.website/",
            "https://www.otherwebsite.com/"
        ]
    }
}
```

Since these noun-verb look ups will happen on every request it makes it a constant lookup on each request instead of scanning db.

Currently I have implement DID.CREATE and CREDENTIAL.CREATE, CREDENTIAL.DELETE

Here is the output from a small test nodejs service that recieves the webhook fired by the SSI-Service:
```
{
  noun: 'DID',
  verb: 'Create',
  url: 'http://host.docker.internal:4321/',
  data: {
    did: {
      '@context': 'https://www.w3.org/ns/did/v1',
      id: 'did:key:z6Mkt11tgxpEBtFPde1T2nBZ7YzR21FejGfHKWEdtRc4MW57',
      verificationMethod: [Array],
      authentication: [Array],
      assertionMethod: [Array],
      keyAgreement: [Array],
      capabilityDelegation: [Array]
    },
    privateKeyBase58: '3cQuZzFFnv2By6VyZvihmDpGLjsEBBdZpnapGuRjEPXV7JppPvFquVVNzvESvfqMtcSaocLorAeShyEvtM5Fk729',
    keyType: 'Ed25519'
  }
}
```


pkg/server/router/credential.go
```diff
@@ -122,6 +123,8 @@ func (cr CredentialRouter) CreateCredential(ctx context.Context, w http.Response
 	}
 
 	resp := CreateCredentialResponse{Credential: createCredentialResponse.Credential, CredentialJWT: createCredentialResponse.CredentialJWT}
+
+	go cr.service.Webhook.PublishWebhook(webhook.Credential, webhook.Create, resp)
```
andresuribe87, line 127: This approach couples the business logic of every endpoint with webhook concerns. 

An alternative that separate these concerns is to have a middleware that is configured per endpoint. Some added benefits include:
* You don't need to update the files in the `router` package.
* Routers do not need to be aware of the `webhook.Service` class.
* It's easy to turn off webhooks via configs.
* More easily manage concurrency (vs spinning a goroutine per request).

nitro-neal: Sounds good! I will look into the middleware solution.

* More easily manage concurrency (vs spinning a goroutine per request).
On this point ^ I think it will still basically spin up a goroutine per request, but I can add a new context and time it out / maybe do some caching so I don't have to access redis so much

andresuribe87: Yes, that definitely works. I meant that you could more easily change that model. 

An alternative model is to have every request write to a channel, and have a pool of goroutines read from them and send the outward events.

pkg/server/middleware/webhook.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/tbd54566975/ssi-service/pkg/server/framework"
+	svcframework "github.com/tbd54566975/ssi-service/pkg/service/framework"
+	"github.com/tbd54566975/ssi-service/pkg/service/webhook"
+)
+
+const (
+	V1Prefix          = "/v1"
+	DIDsPrefix        = "/dids"
+	CredentialsPrefix = "/credentials"
+)
```
decentralgabe, line 18: do we need to re-define these?

pkg/service/webhook/model.go
```diff
@@ -1,45 +1,57 @@
...
 
+import (
+	"fmt"
+	"net/url"
+)
+
 // In the context of webhooks, it's common to use noun.verb notation to describe events,
 // such as "credential.create" or "schema.delete".
 type Noun string
 type Verb string
```
decentralgabe, line 11: there's a risk here that noun and verb have whitespace or paths

meaning we need to make sure they're URL-safe

pkg/service/webhook/model.go
```diff
@@ -58,3 +72,40 @@ type GetSupportedNounsResponse struct {
...
+
+func (cwr CreateWebhookRequest) IsValid() bool {
+	fmt.Println(cwr)
+	if isValidNoun(cwr.Noun) && isValidVerb(cwr.Verb) && isValidURL(cwr.URL) {
+		return true
+	}
+	return false
+}
+
+func isValidNoun(noun Noun) bool {
```
decentralgabe, line 92: instead this can be a receiver method such as

`func (n Noun) IsValid() bool { ... }`

same for below

pkg/service/webhook/model.go
nitro-neal: not sure what you mean by 'URL' safe, these are strings not urls defined by us:
nouns:
credential
schema
did


verbs:
create
delete


they would run getsupporrtednouns and getsupportedverbs, find the strings they want that makes a valid noun.verb pairing (otherwise we wont accept and return error)

nitro-neal:
Updated so that a new webhook middleware is configured on a per route basis:

Example Flow:
Create new webhooks 

PUT - http://localhost:8080/v1/webhooks
```
{
  "noun": "DID",
  "verb": "Create",
  "url": "http://host.docker.internal:4321/"
}
```

```
{
  "noun": "DID",
  "verb": "Delete",
  "url": "http://host.docker.internal:4321/"
}
```

This basically is configuring the SSI-Service to post the response body to the configured url (http://host.docker.internal:4321/) when one of these events happen. (In this case when a DID is created or deleted)

now when someone creates a did:

PUT - http://localhost:8080/v1/dids/:method

My test nodejs service that is accepting requests from http://host.docker.internal:4321/ gets a POST from the SSI-Service in the form of:

```
{
  noun: 'DID',
  verb: 'Create',
  url: 'http://host.docker.internal:4321/',
  data: '{"did":{"@context":"https://www.w3.org/ns/did/v1","id":"did:key:z6MkuN2g4DAEYjnTQwHYMYftszfFRJrTK9rxzHeehAYogZDb","verificationMethod":[{"id":"#z6MkuN2g4DAEYjnTQwHYMYftszfFRJrTK9rxzHeehAYogZDb","type":"Ed25519VerificationKey2018","controller":"did:key:z6MkuN2g4DAEYjnTQwHYMYftszfFRJrTK9rxzHeehAYogZDb","publicKeyBase58":"FumdTxuoDCHzJSSqfyi42u7FbjabuGccJGjirtanmLSD"}],"authentication":[["#z6MkuN2g4DAEYjnTQwHYMYftszfFRJrTK9rxzHeehAYogZDb"]],"assertionMethod":[["#z6MkuN2g4DAEYjnTQwHYMYftszfFRJrTK9rxzHeehAYogZDb"]],"keyAgreement":[["#z6MkuN2g4DAEYjnTQwHYMYftszfFRJrTK9rxzHeehAYogZDb"]],"capabilityDelegation":[["#z6MkuN2g4DAEYjnTQwHYMYftszfFRJrTK9rxzHeehAYogZDb"]]},"privateKeyBase58":"3p6wbbJ3egoDXreF3MRsRWBNUguFoKSakxWdUgb2w8JsCnqZuZT1Nv5kT2EopHoqjEmQHqPTXBQjPgJbJUxZY5ub","keyType":"Ed25519"}'
}
```

pkg/service/webhook/service.go
```diff
@@ -83,3 +163,56 @@ func (s Service) GetSupportedNouns() GetSupportedNounsResponse {
...
+	}
+
+	if is2xxResponse(resp.StatusCode) {
+		return "", fmt.Errorf("status code %v not in the 200s. body: %s", resp.StatusCode, string(body))
+	}
+
+	return string(body), err
+}
+
+func is2xxResponse(statusCode int) bool {
```
decentralgabe, line 216: I saw @andresuribe87 make a change around this recently, is it in a common util package we could pull from?

pkg/service/webhook/storage.go
```diff
@@ -18,18 +23,54 @@ func NewWebhookStorage(db storage.ServiceStorage) (*Storage, error) {
...
 
-func (whs *Storage) GetWebhook(ctx context.Context, id string) (*Webhook, error) {
-	return nil, nil
+func (whs *Storage) GetWebhook(ctx context.Context, noun, verb string) (*Webhook, error) {
+	webhookBytes, err := whs.db.Read(ctx, webhookNamespace, getWebhookKey(noun, verb))
+	if err != nil {
+		return nil, util.LoggingErrorMsg(err, "read db")
+	}
+
+	if webhookBytes == nil || len(webhookBytes) == 0 {
```
decentralgabe, line 40: it might be better to do the unmarshal first, and define an `isEmpty` method on the webhook struct which you call after

decentralgabe (approved):
awesome work here this is huge.

should save merge for final review from @andresuribe87

pkg/service/webhook/model.go
decentralgabe: ah my misunderstanding

pkg/service/webhook/storage.go
```diff
@@ -18,18 +23,54 @@ func NewWebhookStorage(db storage.ServiceStorage) (*Storage, error) {
...
+}
+
+func (whs *Storage) DeleteWebhook(ctx context.Context, noun, verb string) error {
+	return whs.db.Delete(ctx, webhookNamespace, getWebhookKey(noun, verb))
 }
 
-func (whs *Storage) DeleteWebhook(ctx context.Context, id string) error {
-	return nil
+func getWebhookKey(noun, verb string) string {
+	return noun + "-" + verb
```
andresuribe87, line 75: Redis keys are typically separated with `:`. 

I would suggest using the following in this case. A nice side benefit is that redis commander understands that separator easily. 
```suggestion
	return "noun:" + noun + ":verb:" + verb
```

pkg/server/middleware/webhook.go
```diff
@@ -0,0 +1,46 @@
...
+type responseWriter struct {
+	http.ResponseWriter
+	buf *bytes.Buffer
+}
+
+func (rw *responseWriter) Write(b []byte) (int, error) {
+	// Write the response body to both the original response writer and the buffer
+	n, err := rw.ResponseWriter.Write(b)
+	if err == nil {
+		rw.buf.Write(b)
```
andresuribe87, line 23: Should this also do error checking?

andresuribe87 (commented):
This is beautiful! A bunch of small suggestion, nothing big. Love the middleware approach and tests. Thanks for adding this!

pkg/server/middleware/webhook.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/tbd54566975/ssi-service/pkg/server/framework"
+	svcframework "github.com/tbd54566975/ssi-service/pkg/service/framework"
+	"github.com/tbd54566975/ssi-service/pkg/service/webhook"
+)
+
+// Define a custom response writer that wraps the original response writer and writes to a buffer
+type responseWriter struct {
+	http.ResponseWriter
+	buf *bytes.Buffer
```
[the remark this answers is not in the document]
nitro-neal: I don't think it would add anything, we can wrap the response body with a new field if needed. I worry if we include the headers we could open ourselves up to security risk like auth tokens in the headers or something

pkg/server/router/webhook.go
```diff
@@ -35,12 +35,11 @@ type CreateWebhookRequest struct {
 	Noun webhook.Noun `json:"noun" validate:"required"`
 	// The verb for the new webhook.eg: Create
 	Verb webhook.Verb `json:"verb" validate:"required"`
-	// The URLS to post the output of this request to Noun.Verb action to.
-	URLS []string `json:"urls" validate:"required"`
+	// The URL to post the output of this request to Noun.Verb action to.
+	URL string `json:"url" validate:"required"`
```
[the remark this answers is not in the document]
nitro-neal: It felt strange having a CreateWebhook route being able to create multiple webhooks urls (almost like we would make a new route called CreateWebhooks)

Also a bit more complexity on deduping urls.
 
But I can add it in if you think it would be beneficial and used like this

pkg/service/webhook/service.go
```diff
@@ -83,3 +163,56 @@ func (s Service) GetSupportedNouns() GetSupportedNounsResponse {
...
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+
+	resp, err := s.httpClient.Do(req)
+	if err != nil {
+		return "", errors.Wrap(err, "client http client")
+	}
+
+	body, err := io.ReadAll(resp.Body)
```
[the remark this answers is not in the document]
nitro-neal: That is true, it is only used in the verbose error if it is not a 2xx, 
I updated it read the body only if in the error if statement to report it to logs

pkg/service/webhook/service.go
```diff
@@ -83,3 +163,56 @@ func (s Service) GetSupportedNouns() GetSupportedNounsResponse {
...
+		}
+		_, err = s.post(url, string(postJSONData))
+		if err != nil {
+			logrus.Warnf("posting payload to %s", url)
+		}
+	}
+}
+
+func (s Service) post(url string, json string) (string, error) {
+	req, err := http.NewRequest(http.MethodPost, url, bytes.NewBuffer([]byte(json)))
```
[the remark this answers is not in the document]
nitro-neal: I did this on purpose to not mess up the context for our DB and stuff.

If this hangs or something with the existing context and we hit unknown services, could this have an affect on our db/routes/timeouts?

pkg/service/webhook/service.go
```diff
@@ -83,3 +163,56 @@ func (s Service) GetSupportedNouns() GetSupportedNounsResponse {
...
+	webhook, err := s.storage.GetWebhook(context.Background(), string(noun), string(verb))
+	if err != nil {
+		logrus.WithError(err).Warn("get webhook")
+	}
+
+	if webhook == nil {
+		return
+	}
+
+	for _, url := range webhook.URLS {
```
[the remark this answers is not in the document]
nitro-neal: Yup! If we have repeated failed requests we would like disable that webhook endpoint, look at metrics lots of stuff

pkg/service/webhook/service.go
```diff
@@ -83,3 +163,56 @@ func (s Service) GetSupportedNouns() GetSupportedNounsResponse {
...
+		postPayload := Payload{Noun: noun, Verb: verb, URL: url, Data: payload}
+
+		postJSONData, err := json.Marshal(postPayload)
+		if err != nil {
+			logrus.Warn("marshal payload")
```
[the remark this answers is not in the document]
nitro-neal: yup good catch, added continue

pkg/server/router/webhook.go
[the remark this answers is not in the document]
andresuribe87: I think either way is fine. Was just curious. I agree that this seems more natural.

pkg/service/webhook/service.go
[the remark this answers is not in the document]
andresuribe87: It wouldn't have an effect since: a) contexts are copied and; b) this is all running in a separate goroutine.

You could either pass in the same context (which ends up being copied), or create a new context with a new deadline (if you want to increase it).

Issue #237, opened 2023-01-05
[Idea] Caching and Configurable DID Resolver


decentralgabe:
As per discussion on https://github.com/TBD54566975/ssi-sdk/issues/38

The DID resolver should
- Support caching with sensible TTLs, perhaps per-method TTLs (e.g. did:key won't need to change, did:web might...)
- Able to integrate external DID resolver(s) with fallbacks?

Issue #29, opened 2022-04-18
Explore integration with the Universal Resolver


decentralgabe:
https://uniissuer.io/
https://uniregistrar.io/
https://univerifier.io/
https://uniresolver.io/

decentralgabe:
Partially handled in #121

andresuribe87:
Done in #314

PR #314, opened 2023-03-08
Adding support for resolving ION dids

andresuribe87:
# Overview
This adds universal resolver for DIDs that use `ion` did method.

# Description
I added the universal resolver docker container. It's called when `ResolveDID` endpoint is called with the registered universal resolver did method. 

Additionally, I changed all the piping so it can be used for verification. 

# How Has This Been Tested?
Locally. Integration tests don't pass because of networking snafu. 

# Checklist

## References
See https://github.com/decentralized-identity/universal-resolver for how the universal resolver is set up. 

Seems this is also related to #237, though there is no caching here.


michaelneale:
hey this is pretty neat! clever approach

https://hub.docker.com/r/universalregistrar/driver-did-ion/ - is that the image that should be use? (I see e2e failing).

andresuribe87:
@michaelneale the image problem is fixed now. PTAL

michaelneale:
looks good - but I am probably not best reviewier for this.

pkg/service/did/resolve/universal.go
```diff
@@ -0,0 +1,49 @@
...
+type UniversalResolver struct {
+	C   http.Client
+	URL string
+}
+
+type Resolver interface {
+	Resolve(did string, opts ...didsdk.ResolutionOptions) (*didsdk.DIDResolutionResult, error)
+}
+
+func (u UniversalResolver) Resolve(did string, _ ...didsdk.ResolutionOptions) (*didsdk.DIDResolutionResult, error) {
```
nitro-neal, line 22: change this to ur?

pkg/service/did/resolve/universal.go
```diff
@@ -0,0 +1,49 @@
...
+	"io"
+	"net/http"
+
+	didsdk "github.com/TBD54566975/ssi-sdk/did"
+	"github.com/goccy/go-json"
+	"github.com/pkg/errors"
+)
+
+type UniversalResolver struct {
+	C   http.Client
```
nitro-neal, line 14: can we use 'client' instead of C unless this is a golang paradigm or something

nitro-neal (approved):
Awesome work!

pkg/service/did/resolve/universal.go
andresuribe87: Done

pkg/service/did/resolve/universal.go
andresuribe87: Done

pkg/service/did/resolve/universal.go
```diff
@@ -0,0 +1,52 @@
...
+
+	respBody, err := io.ReadAll(bufio.NewReader(resp.Body))
+	if err != nil {
+		return nil, err
+	}
+	var result ResolutionResult
+	err = json.Unmarshal(respBody, &result)
+	if err != nil {
+		return nil, errors.Wrap(err, "unmarshalling JSON")
+	}
```
decentralgabe, line 39:
```suggestion
	if err = json.Unmarshal(respBody, &result); err != nil {
		return nil, errors.Wrap(err, "unmarshalling JSON")
	}
```

pkg/service/did/service.go
```diff
@@ -71,16 +74,29 @@ func NewDIDService(config config.DIDServiceConfig, s storage.ServiceStorage, key
...
+		sdkResolver:      localResolver,
+		methodToResolver: make(map[string]resolve.Resolver),
+	}
+
+	for _, sm := range service.sdkResolver.SupportedMethods() {
+		service.methodToResolver[sm.String()] = service.sdkResolver
+	}
+
+	ur := &resolve.UniversalResolver{
+		Client: http.Client{},
```
decentralgabe, line 95: are there any default params we should set here - like timeouts, etc?

pkg/service/did/service.go
```diff
@@ -119,9 +135,9 @@ func (s *Service) ResolveDID(request ResolveDIDRequest) (*ResolveDIDResponse, er
 	if request.DID == "" {
 		return nil, util.LoggingNewError("cannot resolve empty DID")
 	}
-	resolved, err := s.resolver.Resolve(request.DID)
+	resolved, err := s.Resolve(request.DID)
 	if err != nil {
-		return nil, util.LoggingErrorMsgf(err, "could not resolve DID: %s", request.DID)
+		return nil, err
```
decentralgabe, line 140: does logging happen in the resolver if not found?

decentralgabe (approved):
looks great to me - using the uni resolver is something we've needed for a while #29

few small comments

one big one -- the uniresolver ties us to its own implementation of these drivers, which expands the risk surface of our service. a specific case I'm thinking of - if we wish to publish a driver for our ion node, what does that mean for the universal resolver and for our service?

we could extend the existing ion driver to let you specify a resolver uri for ion, we could submit a competing driver, or we could choose to use our own resolution logic in this service for ion.

something to consider

pkg/service/manifest/service.go
```diff
@@ -72,13 +72,7 @@ func (s Service) Config() config.ManifestServiceConfig {
...
 }
 
-func NewManifestService(
-	config config.ManifestServiceConfig,
-	s storage.ServiceStorage,
-	keyStore *keystore.Service,
-	didResolver *didsdk.Resolver,
-	credential *credential.Service,
-) (*Service, error) {
+func NewManifestService(config config.ManifestServiceConfig, s storage.ServiceStorage, keyStore *keystore.Service, didResolver resolve.Resolver, credential *credential.Service) (*Service, error) {
```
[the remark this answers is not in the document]
andresuribe87: Haha - maybe something for another day. I've have mixed successes with `wire`. 

My biggest concern is that rampup time is quite high for folks that haven't dealt with it.

OTOH, it forces you to use good practices and unit tests are quite simple.

pkg/service/did/service.go
andresuribe87: this is a good point. Instead of defining defaults, I made the changes so that the context is propagated properly. The timeouts will be inherited from the context. This should also give us spans and traces for free.

pkg/service/did/service.go
```diff
@@ -3,23 +3,26 @@ package did
...
 	"github.com/tbd54566975/ssi-service/pkg/storage"
 )
 
 type Service struct {
-	config   config.DIDServiceConfig
-	storage  *Storage
-	resolver *didsdk.Resolver
+	config      config.DIDServiceConfig
+	storage     *Storage
+	sdkResolver *didsdk.Resolver
```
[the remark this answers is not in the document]
andresuribe87: Did this. Also had to create a LocalResolver wrapper because the SDK implementation does not accept contexts (which can arguably be part of separate issue).

pkg/service/did/resolve/universal.go
andresuribe87: Done

PR #321, opened 2023-03-10
Test with TBD stuff

andresuribe87:
This also should fix the flakiness in #319

PR #318, opened 2023-03-09
Fix concurrency issues

andresuribe87:
# Overview
This fixes #302 . See the issue details for the root cause.

# Description
This PR mainly moves the call to `GetStatusListCredentialKeyData` so that it's inside the Optimistic Concurrency block.
Additionally, this PR modifies the test so that networking failures do not fail it.
Finally, the test was extended to catch the problem described in https://github.com/TBD54566975/ssi-service/issues/302#issuecomment-1459104587

# How Has This Been Tested?
See updated test.

# Checklist

Before submitting this PR, please make sure:

- [x] I have read the CONTRIBUTING document.
- [x] My code is consistent with the rest of the project 
- [x] I have tagged the relevant reviewers and/or interested parties
- [x] I have updated the READMEs and other documentation of affected packages

## References
N/A

PR #323, opened 2023-03-13
Combining all changes that fix flakiness

andresuribe87:
This change combines https://github.com/TBD54566975/ssi-service/pull/321 and https://github.com/TBD54566975/ssi-service/pull/318 because they were both causes for flakiness in e2e tests. Refer to those PRs if you want to see isolated changes.

### DIFF
--- a/integration/credential_revocation_concurrency_integration_test.go
+++ b/integration/credential_revocation_concurrency_integration_test.go
@@ -111,17 +111,25 @@ func TestConcurrencyRevocationVerifiableCredentialIntegration(t *testing.T) {
 	assert.NotEmpty(t, schemaID)
 
 	const vcCount = 100
-	credStatusListIndexes := make([]string, vcCount)
+	credStatusListIndexes := make([]string, 0, vcCount)
 
 	var wg sync.WaitGroup
 	wg.Add(vcCount)
 
+	m := sync.Mutex{}
+	credStatusListURLSet := make(map[string]struct{})
+
 	for i := 0; i < vcCount; i++ {
-		go func(i int) {
+		go func() {
 			defer wg.Done()
 
 			vcOutput, err := CreateVerifiableCredential(credInputParams{IssuerID: issuerDID, SchemaID: schemaID, SubjectID: issuerDID}, true)
-			assert.NoError(t, err)
+
+			// We're hammering the DB, so some calls might fail due to internal timeouts or similar. Upon failure, we
+			// shouldn't check any assertions, since we know they'll fail.
+			if err != nil {
+				return
+			}
 			assert.NotEmpty(t, vcOutput)
 
 			credStatusURL, err := getJSONElement(vcOutput, "$.credential.credentialStatus.id")
@@ -138,13 +146,17 @@ func TestConcurrencyRevocationVerifiableCredentialIntegration(t *testing.T) {
 			assert.NoError(t, err)
 			assert.NotEmpty(t, credStatusListIndex)
 
-			credStatusListIndexes[i] = credStatusListIndex
-		}(i)
+			m.Lock()
+			credStatusListIndexes = append(credStatusListIndexes, credStatusListIndex)
+			credStatusListURLSet[credStatusListURL] = struct{}{}
+			m.Unlock()
+		}()
 	}
 
 	wg.Wait()
 
-	assert.True(t, areElementsUnique(credStatusListIndexes))
+	assert.Len(t, credStatusListURLSet, 1)
+	assert.True(t, areElementsUnique(credStatusListIndexes), "elements should be unique")
 }
 
 func areElementsUnique(arr []string) bool {
--- a/pkg/service/credential/storage.go
+++ b/pkg/service/credential/storage.go
@@ -11,7 +11,6 @@ import (
 	"github.com/TBD54566975/ssi-sdk/credential/signing"
 	statussdk "github.com/TBD54566975/ssi-sdk/credential/status"
 	"github.com/goccy/go-json"
-	"github.com/google/uuid"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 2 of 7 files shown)


pkg/testutil/well_known_generation_test.go
```diff
@@ -72,7 +72,7 @@ func createWellKnownDIDConfiguration(tt *testing.T, didResponse *did.CreateDIDRe
 	err = builder.SetCredentialSubject(subject)
 	assert.NoError(tt, err)
 
-	err = builder.AddContext("https://identity.foundation/.well-known/did-configuration/v1")
+	err = builder.AddContext("https://tbd.website/.well-known/did-configuration/v1")
```
nitro-neal, line 75: love it

pkg/testutil/well_known_generation_test.go
```diff
@@ -137,7 +137,7 @@ func createWellKnownDIDConfiguration(tt *testing.T, didResponse *did.CreateDIDRe
 	credToken := keyaccess.JWT(signedTokenBytes).Ptr()
 
 	wellKnown := map[string]interface{}{
-		"@context":    "https://identity.foundation/.well-known/did-configuration/v1",
+		"@context":    "https://tbd.website/.well-known/did-configuration/v1",
```
nitro-neal, line 140: this is 404 btw

nitro-neal (approved):
Awesome thanks for doing this!

pkg/service/credential/storage.go
```diff
@@ -623,29 +616,24 @@ func (cs *Storage) GetStatusListCurrentIndexWatchKey(issuer, schema, statusPurpo
...
 	storedStatusListCreds, err := cs.GetStatusListCredentialsByIssuerSchemaPurpose(ctx, issuer, schema, statusPurpose)
 	if err != nil {
-		return "", nil, util.LoggingNewErrorf("getting status list credential for issuer: %s schema: %s", issuer, schema)
+		return nil, util.LoggingNewErrorf("getting status list credential for issuer: %s schema: %s", issuer, schema)
 	}
 
 	// This should never happen, there should always be only 1 status list credential per <issuer,schema, statusPurpose> triplet
 	if len(storedStatusListCreds) > 1 {
-		return "", nil, util.LoggingNewErrorf("only one status list credential per <issuer,schema> pair allowed. issuer: %s schema: %s", issuer, schema)
+		return nil, util.LoggingNewErrorf("only one status list credential per <issuer,schema> pair allowed. issuer: %s schema: %s", issuer, schema)
```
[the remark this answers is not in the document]
andresuribe87: I'm not sure if the error you were seeing was a root cause, or a side effect.

We [should not panic](https://github.com/uber-go/guide/blob/master/style.md#dont-panic). In this case, if this situation is truly a bad state, we should emit a FATAL error. I don't see this warranting stopping the execution of ssi-service. Happy to talk through this if you think we should!

pkg/testutil/well_known_generation_test.go
andresuribe87: Good point. I'm going to leave as is. This file seems to be ignored in tests, and appears to be meant as an example only.

integration/credential_revocation_concurrency_integration_test.go
```diff
@@ -111,17 +111,25 @@ func TestConcurrencyRevocationVerifiableCredentialIntegration(t *testing.T) {
 	assert.NotEmpty(t, schemaID)
 
 	const vcCount = 100
-	credStatusListIndexes := make([]string, vcCount)
+	credStatusListIndexes := make([]string, 0, vcCount)
 
 	var wg sync.WaitGroup
 	wg.Add(vcCount)
 
+	m := sync.Mutex{}
```
decentralgabe, line 119: var m sync.Mutex

integration/credential_revocation_concurrency_integration_test.go
```diff
@@ -111,17 +111,25 @@ func TestConcurrencyRevocationVerifiableCredentialIntegration(t *testing.T) {
 	assert.NotEmpty(t, schemaID)
 
 	const vcCount = 100
-	credStatusListIndexes := make([]string, vcCount)
+	credStatusListIndexes := make([]string, 0, vcCount)
```
decentralgabe, line 114: if 0 init does this syntax give us a benefit over `var credStatusListIndexes []string`?

pkg/service/credential/storage.go
```diff
@@ -623,29 +616,24 @@ func (cs *Storage) GetStatusListCurrentIndexWatchKey(issuer, schema, statusPurpo
...
 	}
 
 	// No Status List Credential Exists, create a new uuid
 	if storedStatusListCreds == nil || len(storedStatusListCreds) == 0 {
 		logrus.Warn("No Status List Credential Exists, create a new uuid")
```
decentralgabe, line 632: would be useful to log the schema and issuer here

pkg/testutil/well_known_generation_test.go
decentralgabe, line 75: test util is for integration tests only? if not we should rethink this

integration/credential_revocation_concurrency_integration_test.go
andresuribe87: Yes, the correct capacity is allocated. More details [here](https://github.com/uber-go/guide/blob/master/style.md#specifying-slice-capacity)

pkg/testutil/well_known_generation_test.go
andresuribe87: testutil is currently being skipped right now. I'm of the opinion that we should remove, and add some of this functionality to a CLI.

integration/credential_revocation_concurrency_integration_test.go
andresuribe87: Adding in another PR.

PR #325, opened 2023-03-14
This time for realz - fix flaky e2e tests

### LINKED ISSUE #302: [Bug] End2End Concurrency Test Issue
**Describe the bug**
End to end test is failing on the concurrency test

**To Reproduce**

Our test:
https://github.com/TBD54566975/ssi-service/blob/main/integration/credential_revocation_concurrency_integration_test.go

is failing with:
```
build-web-1         | {"level":"info","msg":"214e6a70-67fd-42b7-ab45-8841311051ab : ERROR : could not create credential: execute: failed to execute after retrying: problem with getting status list credential: only one status list credential per \u003cissuer,schema\u003e pair allowed. issuer: did:key:z6Mkqgq6tynyUAvxftZCR65exq8gzwSeKfUCDYtT4cY8FR6h schema: 72fcd4fd-3e60-4af3-af20-9b7600491f3f","time":"2023-03-01T18:46:55Z"}
```

We should never be in this state, with the watch keys watching when a status list credential gets created. 
Definitely need to debug and fix our concurrency issue here

andresuribe87:
I believe this is the final nail in the coffin in order to fully fix #302. The actual fix is really simple: return an error when there is a read error. It was tough to spot, but I have a high degree of confidence that this is it.

Separately, I included the following changes that aided the investigation:
* Uploading a logs artifact when the e2e test fails.
* Only startup the set of docker containers that's actually needed (this might be debatable, but makes the docker setup step much faster).

### DIFF
--- a/pkg/service/credential/storage.go
+++ b/pkg/service/credential/storage.go
@@ -504,19 +504,20 @@ func (cs *Storage) GetStatusListCredentialsByIssuerSchemaPurpose(ctx context.Con
 	}
 
 	// now get each credential by key
-	var storedCreds []StoredCredential
+	storedCreds := make([]StoredCredential, 0, len(issuerSchemaKeys))
 	for _, key := range issuerSchemaKeys {
 		credBytes, err := cs.db.Read(ctx, statusListCredentialNamespace, key)
 		if err != nil {
 			logrus.WithError(err).Errorf("could not read credential with key: %s", key)
-		} else {
-			var cred StoredCredential
-			if err = json.Unmarshal(credBytes, &cred); err != nil {
-				logrus.WithError(err).Errorf("unmarshalling credential with key: %s", key)
-			}
-
-			storedCreds = append(storedCreds, cred)
+			return nil, err
 		}
+
+		var cred StoredCredential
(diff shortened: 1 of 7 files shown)


pkg/service/credential/storage.go
```diff
@@ -504,19 +504,20 @@ func (cs *Storage) GetStatusListCredentialsByIssuerSchemaPurpose(ctx context.Con
...
 	for _, key := range issuerSchemaKeys {
 		credBytes, err := cs.db.Read(ctx, statusListCredentialNamespace, key)
 		if err != nil {
 			logrus.WithError(err).Errorf("could not read credential with key: %s", key)
-		} else {
-			var cred StoredCredential
-			if err = json.Unmarshal(credBytes, &cred); err != nil {
-				logrus.WithError(err).Errorf("unmarshalling credential with key: %s", key)
-			}
+			return nil, err
```
andresuribe87, line 512: This wasn't erroring out. Which ended up causing some bizarre failures in the e2e tests.

nitro-neal: good find

nitro-neal (approved):
Great job finding this

PR #326, opened 2023-03-14
Enable open telemetry tracing for redis

andresuribe87:
Also made a couple of simple changes so tracing runs locally (whether running with docker or in the local machine). See below for an example. Redis was also instrumented. 

image


decentralgabe (approved):
🚢

PR #330, opened 2023-03-17
Host Credential Issuer Metadata in a separate executable

andresuribe87:
copilot:all

# Overview
This PR does two things: 
2. Hosts the credential issuer metadata in `/.well-known/openid-credential-issuer`. 
3. Enabled users of the service to specify a path where they want to have their own custom metadata.

# Description
I intentionally tried to keep this simple.

# How Has This Been Tested?
Added a simple unit test that fetches from the endpoint.

# Checklist

## References
https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-credential-issuer-metadata-


pkg/authorizationserver/oauth2.go
```diff
@@ -0,0 +1,138 @@
...
+//  2. store - no auth service is generally useful unless it can remember clients and users.
+//     fosite is incredibly composable, and the store parameter enables you to build and BYODb (Bring Your Own Database)
+//  3. secret - required for code, access and refresh token generation.
+//  4. privateKey - required for id/jwt token generation.
+var (
+	// Check the api documentation of `compose.Config` for further configuration options.
+	fositeConfig = &fosite.Config{
+		AccessTokenLifespan: time.Minute * 30,
+		GlobalSecret:        secret,
+		// ...
```
nitro-neal, line 32: Is this an official oatuh2.go file so you left as is? Otherwise can remove this comment

pkg/authorizationserver/oauth2_auth.go
```diff
@@ -0,0 +1,98 @@
...
+		requestedScopes += fmt.Sprintf(`<li><input type="checkbox" name="scopes" value="%s">%s</li>`, this, this)
+	}
+
+	// Normally, this would be the place where you would check if the user is logged in and gives his consent.
+	// We're simplifying things and just checking if the request includes a valid username and password
+	if err := req.ParseForm(); err != nil {
+		oauth2.WriteAuthorizeError(ctx, rw, ar, err)
+		return nil
+	}
+	if req.PostForm.Get("username") != "peter" {
```
nitro-neal, line 34: did you mean to leave this?

pkg/authorizationserver/oauth2.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/ory/fosite/compose"
+	"github.com/ory/fosite/handler/openid"
+	"github.com/ory/fosite/storage"
+	"github.com/ory/fosite/token/jwt"
+	"github.com/tbd54566975/ssi-service/config"
+	"github.com/tbd54566975/ssi-service/pkg/server/framework"
+	"github.com/tbd54566975/ssi-service/pkg/server/middleware"
+)
+
+// fosite requires four parameters for the server to get up and running:
```
nitro-neal, line 21: Are these official oauth files and we should't change them?

pkg/authorizationserver/oauth2.go
```diff
@@ -0,0 +1,138 @@
...
+	//	if err != nil {
+	//		panic(err)
+	//	}
+	// }
+	// ```
+	//
+	// If you require this to key to be stable, for example, when running multiple fosite servers, you can generate the
+	// 32byte random key as above and push it out to a base64 encoded string.
+	// This can then be injected and decoded as the `var secret []byte` on server start.
+	secret = []byte("some-cool-secret-that-is-32bytes")
```
decentralgabe, line 67: can read this from config

decentralgabe: or env var

decentralgabe (approved):

michaelneale:
copilot:summary

nitro-neal (approved):

pkg/authorizationserver/issuermetadata.go
```diff
@@ -0,0 +1,28 @@
...
+
+	"github.com/TBD54566975/ssi-sdk/oidc/issuance"
+	"github.com/goccy/go-json"
+	"github.com/tbd54566975/ssi-service/pkg/server/framework"
+)
+
+// Allow retrieval of credential issuer metadata according to https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-credential-issuer-metadata
+func credentialIssuerMetadata(config *AuthConfig) func(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
+	// It's ok to panic inside this function because this is only called during startup.
+	jsonData, err := os.ReadFile(config.CredentialIssuerFile)
```
nitro-neal, line 16: if this is the same file being retrieved over and over we can cache it in memory?

andresuribe87: It is actually cached! This function is only run once, during the setup of the server. If you look closely, this function returns _another function_. The latter uses the already read file.

PR #333, opened 2023-03-20
Adding functionality for soft deletion of dids

nitro-neal:
# Overview
Adding functionality for soft deletion of dids

# Description
This change adds a new endpoint: 
DEL - http://localhost:8080/v1/dids/:method/:id
When this is called with the correct did method and id it will flip the softDelete flag to true for the db entry.

A user can still get the did if they know the DID ID, 
GET - http://localhost:8080/v1/dids/:method/:id ✅ 

But when they get all dids this will not appear
GET - http://localhost:8080/v1/dids/:method ⚠️ 

The reason for this change is to facilitate a clean SSI-Service Admin UI but not leave any hanging VCs with unaccessible DIDS.

### DIFF
--- a/pkg/server/router/did.go
+++ b/pkg/server/router/did.go
@@ -222,6 +222,52 @@ type ResolveDIDResponse struct {
 	DIDDocumentMetadata *didsdk.DIDDocumentMetadata   `json:"didDocumentMetadata,omitempty"`
 }
 
+type DeleteDIDByMethodRequest struct {
+	// Identifies the cryptographic algorithm family to use when generating this key.
+	// One of the following: `"Ed25519","X25519","secp256k1","P-224","P-256","P-384","P-521","RSA"`.
+	KeyType crypto.KeyType `json:"keyType" validate:"required"`
+
+	// Required when creating a DID with the `web` did method. E.g. `did:web:identity.foundation`.
+	DIDWebID string `json:"didWebId"`
+}
+
+// SoftDeleteDIDByMethod godoc
+//
+// @Summary     Soft Delete DID
+// @Description Soft Deletes DID by method
+// @Tags        DecentralizedIdentityAPI
+// @Accept      json
+// @Produce     json
+// @Param       request body     DeleteDIDByMethodRequest true "request body"
+// @Param       method  path     string                   true "Method"
+// @Param       id      path     string                   true "ID"
+// @Success     200     {string} string "OK"
+// @Failure     400     {string} string "Bad request"
+// @Router      /v1/dids/{method}/{id} [delete]
+func (dr DIDRouter) SoftDeleteDIDByMethod(ctx context.Context, w http.ResponseWriter, _ *http.Request) error {
+	method := framework.GetParam(ctx, MethodParam)
+	if method == nil {
+		errMsg := "soft delete DID by method request missing method parameter"
+		logrus.Error(errMsg)
+		return framework.NewRequestErrorMsg(errMsg, http.StatusBadRequest)
+	}
+	id := framework.GetParam(ctx, IDParam)
+	if id == nil {
+		errMsg := fmt.Sprintf("soft delete DID request missing id parameter for method: %s", *method)
+		logrus.Error(errMsg)
+		return framework.NewRequestErrorMsg(errMsg, http.StatusBadRequest)
+	}
+
+	deleteDIDRequest := did.DeleteDIDRequest{Method: didsdk.Method(*method), ID: *id}
+	if err := dr.service.SoftDeleteDIDByMethod(ctx, deleteDIDRequest); err != nil {
+		errMsg := fmt.Sprintf("could not soft delete DID with id: %s", *id)
+		logrus.WithError(err).Error(errMsg)
+		return framework.NewRequestError(errors.Wrap(err, errMsg), http.StatusInternalServerError)
+	}
+
+	return framework.Respond(ctx, w, nil, http.StatusOK)
+}
+
 // ResolveDID godoc
 //
 // @Summary     Resolve a DID
--- a/pkg/server/router/did.go
+++ b/pkg/server/router/did.go
@@ -222,27 +222,20 @@ type ResolveDIDResponse struct {
 	DIDDocumentMetadata *didsdk.DIDDocumentMetadata   `json:"didDocumentMetadata,omitempty"`
 }
 
-type DeleteDIDByMethodRequest struct {
-	// Identifies the cryptographic algorithm family to use when generating this key.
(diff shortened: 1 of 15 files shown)


pkg/server/router/did.go
```diff
@@ -222,6 +222,52 @@ type ResolveDIDResponse struct {
 	DIDDocumentMetadata *didsdk.DIDDocumentMetadata   `json:"didDocumentMetadata,omitempty"`
 }
 
+type DeleteDIDByMethodRequest struct {
+	// Identifies the cryptographic algorithm family to use when generating this key.
+	// One of the following: `"Ed25519","X25519","secp256k1","P-224","P-256","P-384","P-521","RSA"`.
+	KeyType crypto.KeyType `json:"keyType" validate:"required"`
+
+	// Required when creating a DID with the `web` did method. E.g. `did:web:identity.foundation`.
+	DIDWebID string `json:"didWebId"`
```
decentralgabe, line 231: should this be required too?

pkg/service/did/storage.go
```diff
@@ -27,8 +27,9 @@ var (
 )
 
 type StoredDID struct {
-	ID  string          `json:"id"`
-	DID did.DIDDocument `json:"did"`
+	ID          string          `json:"id"`
+	DID         did.DIDDocument `json:"did"`
+	SoftDeleted bool            `json:"softDeleted"`
```
decentralgabe, line 32: should we add a `DeletedAt` property?

decentralgabe (approved):
lgtm hold for the timestamp comment

pkg/service/did/key.go
```diff
@@ -99,7 +100,26 @@ func (h *keyDIDHandler) GetDIDs(ctx context.Context, method did.Method) (*GetDID
...
+	id := request.ID
+	gotStoredDID, err := h.storage.GetDID(ctx, id)
+	if err != nil {
+		return fmt.Errorf("error getting DID: %s", id)
+	}
+	if gotStoredDID == nil {
+		return fmt.Errorf("did with id<%s> could not be found", id)
+	}
+
+	gotStoredDID.SoftDeleted = true
```
andresuribe87, line 122: Do we need to do optimistic locking for this update? Is it possible to modify an already stored did within out system?

pkg/server/router/did.go
nitro-neal: If we had it required then when you would make a did-key you would have to have didWebId set to blank (or anything)

Definitely should change this, would require some refactoring on how we handle the did creation endpoint

pkg/server/router/did.go
```diff
@@ -222,6 +222,52 @@ type ResolveDIDResponse struct {
...
+}
+
+// SoftDeleteDIDByMethod godoc
+//
+// @Summary     Soft Delete DID
+// @Description Soft Deletes DID by method
```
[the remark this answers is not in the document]
nitro-neal: Added:
// @Description When this is called with the correct did method and id it will flip the softDelete flag to true for the db entry.
// @Description A user can still get the did if they know the DID ID, and the did keys will still exist, but this did will not show up in the GetDIDsByMethod call
// @Description This facilitates a clean SSI-Service Admin UI but not leave any hanging VCs with inaccessible hanging DIDs.

pkg/server/router/did.go
```diff
@@ -222,6 +222,52 @@ type ResolveDIDResponse struct {
...
+// @Tags        DecentralizedIdentityAPI
+// @Accept      json
+// @Produce     json
+// @Param       request body     DeleteDIDByMethodRequest true "request body"
+// @Param       method  path     string                   true "Method"
+// @Param       id      path     string                   true "ID"
+// @Success     200     {string} string "OK"
+// @Failure     400     {string} string "Bad request"
```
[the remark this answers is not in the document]
nitro-neal: lol thanks for keeping me honest, I wonder if the linter can help with this stuff?

pkg/server/router/did.go
```diff
@@ -222,6 +222,52 @@ type ResolveDIDResponse struct {
...
+	}
+
+	deleteDIDRequest := did.DeleteDIDRequest{Method: didsdk.Method(*method), ID: *id}
+	if err := dr.service.SoftDeleteDIDByMethod(ctx, deleteDIDRequest); err != nil {
+		errMsg := fmt.Sprintf("could not soft delete DID with id: %s", *id)
+		logrus.WithError(err).Error(errMsg)
+		return framework.NewRequestError(errors.Wrap(err, errMsg), http.StatusInternalServerError)
+	}
+
+	return framework.Respond(ctx, w, nil, http.StatusOK)
```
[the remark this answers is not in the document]
nitro-neal: Added 204 everywhere we do delete 👍

pkg/server/router/did.go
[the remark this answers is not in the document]
nitro-neal: 🤔  I think it should be fine, if you know the DID you can always do a GET so you 'know' that it still exists in there, calling delete again will just flip to true -again- (which does nothing)

pkg/service/did/key.go
nitro-neal: I dont think we need locking here since the flipping of SoftDeleted doesn't rely on any other entity.
Currently we can only create, get and delete dids, so no updating can happen.

pkg/server/router/did.go
decentralgabe: gotcha

I think we could amend this to be a generic "option" field that has unique meaning per DID method

pkg/service/did/key.go
andresuribe87: Will we want to support updating DIDs? For example, what if someone wants to do an ION operation?

Deletion then would compete against an update operation, and we could end up losing some data. That sounds like something we'll be adding soon(ish). So I would safe better to be safe than sorry.

pkg/server/router/did.go
nitro-neal: Added a new issue - https://github.com/TBD54566975/ssi-service/issues/335

pkg/server/router/did.go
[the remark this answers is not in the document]
andresuribe87: I bet we could write a linter for it. Tried quickly with Chad, but couldn't get exactly what we wanted... Maybe for another day :P

pkg/service/did/key.go
nitro-neal: If the ion did get updated in the bitcoin 'database' and ours only points to it I think it will still be fine I think, its only if we do any updates to existing dids in our DB.

I made an issue regardless and yea it couldn't hurt to throw it in there. We should have a bigger store to use our optimistic locking in many places. - https://github.com/TBD54566975/ssi-service/issues/336

PR #334, opened 2023-03-21
Use published images default

michaelneale:
Small change to allow the published images to be used by default when `--build` isn't used for docker compose.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,10 +81,17 @@ the service, you can use the following command, which will start the service on 
 mage run
 ```
 
-Or, you can run docker-compose yourself:
+Or, you can run docker-compose yourself, building from source:
 ```shell
 cd build && docker-compose up --build
 ```
+
+To use the pre-published images:
+```shell
+cd build && docker-compose up -d
+```
+
+
 
 ## Health and Readiness Checks
 
--- a/build/docker-compose.yml
+++ b/build/docker-compose.yml
@@ -3,6 +3,7 @@ version: "3.98"
 services:
   web:
     container_name: web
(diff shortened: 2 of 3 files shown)


build/docker-compose.yml
```diff
@@ -3,6 +3,7 @@ version: "3.98"
 services:
   web:
     container_name: web
+    image: ghcr.io/tbd54566975/ssi-service:main
     build:
```
jcdby, line 7: without deleting the build blocks here, it will still try to build it with the defined dockerfile even you added an "image" here. 
That is the case on my computer, just FYI

michaelneale: hrm - maybe it didn't on mine as it had already built it previously?

michaelneale: ok may just need a different compose file for running it from images

michaelneale:
@jcdby try it now with the new instructions in the README - a separate compose file

jcdby:
thank you, will try it

README.md
```diff
@@ -81,11 +81,18 @@ the service, you can use the following command, which will start the service on
...
 
-Or, you can run docker-compose yourself:
+Or, you can run docker-compose yourself, building from source:
 ```shell
 cd build && docker-compose up --build
 ```
 
+To use the pre-published images:
+```shell
+cd build && docker-compose up -d
```
jcdby, line 91: I think we might change this instruction to "cd build && docker-compose -f docker-compose-published.yml up -d" which will run the new yml file

decentralgabe (approved):

PR #360, opened 2023-03-28
refactoring DID resolution

decentralgabe:
- unified handlers
- removed cycles by having `internal` not import anything from `service` (we should keep this pattern)


pkg/service/did/resolution/resolver.go
```diff
@@ -0,0 +1,107 @@
+package resolution
```
decentralgabe, line 1: interesting stuff here

pkg/service/did/resolution/local.go
```diff
@@ -0,0 +1,29 @@
+package resolution
+
+import (
+	"context"
+
+	didsdk "github.com/TBD54566975/ssi-sdk/did"
+	"github.com/pkg/errors"
+)
+
+// localResolver is an implementation of Resolver that passes through the parameters into the sdk implementation that
```
andresuribe87, line 10:
```suggestion
// localResolver is an implementation of didsdk.Resolution that passes through the parameters into the sdk implementation that
```

pkg/service/did/resolution/resolver.go
```diff
@@ -0,0 +1,108 @@
...
+// 1. Try to resolve with the handlers we have, wrapping the resulting DID in resolution result
+// 2. Try to resolve with the local resolver
+// 3. Try to resolve with the universal resolver
+func (sr *ServiceResolver) Resolve(ctx context.Context, did string, opts ...didsdk.ResolutionOptions) (*didsdk.ResolutionResult, error) {
+	// check the did is valid
+	if _, err := utilint.GetMethodForDID(did); err != nil {
+		return nil, errors.Wrap(err, "getting method DID")
+	}
+
+	ae := util.NewAppendError()
```
andresuribe87, line 71: When I think of errors being returned, I'm typically thinking of what callers of the method can do with them.

Given the above, curious what's the rationale for returning multiple errors in a single error? Could this be simplified by returning a single error?

pkg/service/did/resolution/universal.go
```diff
@@ -11,19 +11,25 @@ import (
 	"github.com/pkg/errors"
 )
 
-// UniversalResolver is a struct that implements the Resolver interface. It calls the universal resolver endpoint
+// universalResolver is a struct that implements the Resolver interface. It calls the universal resolver endpoint
 // to resolve any DID according to https://github.com/decentralized-identity/universal-resolver.
-type UniversalResolver struct {
-	Client http.Client
+type universalResolver struct {
+	Client *http.Client
```
andresuribe87, line 17: You probably want to unexport the fields as well.

pkg/service/did/service.go
```diff
@@ -74,72 +73,41 @@ func NewDIDService(config config.DIDServiceConfig, s storage.ServiceStorage, key
...
+		return nil, errors.New(service.Status().Message)
 	}
-	return nil
+	return &service, nil
 }
 
 func (s *Service) ResolveDID(request ResolveDIDRequest) (*ResolveDIDResponse, error) {
 	if request.DID == "" {
-		return nil, util.LoggingNewError("cannot resolve empty DID")
+		return nil, util.LoggingNewError("cannot resolution empty DID")
```
andresuribe87, line 110:
```suggestion
		return nil, util.LoggingNewError("cannot resolve empty DID")
```

pkg/service/presentation/service.go
```diff
@@ -227,6 +231,27 @@ func (s Service) CreateSubmission(ctx context.Context, request model.CreateSubmi
 	}, nil
 }
 
+// verifyTokenFromDID verifies the token from the did by first resolving the did, next extracting the public key,
+// and then verifying the token
+func (s Service) verifyTokenFromDID(ctx context.Context, did string, token keyaccess.JWT) error {
```
andresuribe87, line 236: This function is repeated. Was this intentional?

pkg/service/did/resolution/resolver.go
```diff
@@ -0,0 +1,108 @@
...
+	return &ServiceResolver{
+		resolutionMethods: resolutionMethods,
+		hr:                handlerResolver,
+		lr:                lr,
+		ur:                ur,
+	}, nil
+}
+
+// Resolve resolves a DID using a combination of local and universal resolvers. The ordering is as follows:
+// 1. Try to resolve with the handlers we have, wrapping the resulting DID in resolution result
```
andresuribe87, line 62: For what use-cases do we need to support resolution with our handlers?

pkg/service/did/handler.go
```diff
@@ -0,0 +1,62 @@
...
+// MethodHandler describes the functionality of *all* possible DID service, regardless of method
+type MethodHandler interface {
+	CreateDID(ctx context.Context, request CreateDIDRequest) (*CreateDIDResponse, error)
+	GetDID(ctx context.Context, request GetDIDRequest) (*GetDIDResponse, error)
+	GetDIDs(ctx context.Context, method didsdk.Method) (*GetDIDsResponse, error)
+	SoftDeleteDID(ctx context.Context, request DeleteDIDRequest) error
+}
+
+// HandlerResolver is a DID resolver that uses MethodHandlers to resolve DIDs, meaning DIDs we store in our database
+type HandlerResolver struct {
```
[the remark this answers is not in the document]
decentralgabe: hmmm...think so, let me try

pkg/service/did/handler.go
```diff
@@ -0,0 +1,62 @@
...
+	switch method {
+	case didsdk.KeyMethod:
+		s.handlers[method] = NewKeyDIDHandler(s.storage, s.keyStore)
+	case didsdk.WebMethod:
+		s.handlers[method] = NewWebDIDHandler(s.storage, s.keyStore)
+	default:
+		return util.LoggingNewErrorf("unsupported DID method: %s", method)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
decentralgabe: yes

pkg/service/did/resolution/resolver.go
decentralgabe: I was thinking that we may not know the exact error to propagate, if multiple resolution methods were attempted and failed. the error should be meaningful, which in this case means showing its lineage

pkg/service/presentation/service.go
decentralgabe: no - will remove one

pkg/jwt/verification.go
```diff
@@ -1,29 +0,0 @@
...
-	didint "github.com/tbd54566975/ssi-service/internal/did"
-	"github.com/tbd54566975/ssi-service/internal/keyaccess"
-	"github.com/tbd54566975/ssi-service/internal/util"
-	"github.com/tbd54566975/ssi-service/pkg/service/did/resolve"
-)
-
-// VerifyTokenFromDID verifies that the information in the token was digitally signed by the public key associated with
-// the public key of the verification method of the did's document. The passed in resolver is used to map from the did
-// to the did document.
-func VerifyTokenFromDID(ctx context.Context, did string, token keyaccess.JWT, s resolve.Resolver) error {
```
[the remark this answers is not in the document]
decentralgabe: I had first put it in internal since it's not tied to any service, but then I saw the dependency on resolver, which is a part of the DID service

then it makes sense that if there's a DID service dependency it shouldn't be in internal...since I had imagined internal only depending on itself

pkg/service/did/resolution/resolver.go
decentralgabe: I am imagining the case of something like did:web or ion - any remote resolution - where we do not need to call out to a service endpoint if we have a local copy. this raises the question of staleness, where we can add a resolution option for `forceRemote`

pkg/service/presentation/service.go
decentralgabe: this is intentional. it is because one is in the presentation service and another in the manifest service. each service has its own dependencies and separate impls, but has two options to verify a token:
1. call the new method in the DID service - `GetKeyFromDID` and verify the token
2. use the resolver interface as it's doing

decentralgabe: an alternative could be to add a new method to the DID service that does this verification

decentralgabe: actually now that the resolver interface is in the SDK I can undo this

pkg/jwt/verification.go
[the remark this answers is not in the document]
decentralgabe: disregard, moved back

pkg/service/did/resolution/resolver.go
andresuribe87: Should we move away from using storage as a DID document cache?

Or if we're using it as a cache, then I would recommend we keep that in memory (or in a redis implementation that doesn't need to guarantee durability).

Why I'm more favorable to removing is because caches are used as a performance optimization, and can lead to staleness bugs. Since we haven't diagnosed performance bottlenecks, I don't think they're worth the complexity. I've already seen some confusion pop up from how the DIDAPI endpoint works.

What are your thoughts? If you agree, mind creating a tracking issue?

andresuribe87 (approved):

pkg/service/did/handler.go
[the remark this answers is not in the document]
decentralgabe: can remove

pkg/service/did/resolution/local.go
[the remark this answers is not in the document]
decentralgabe: yes

pkg/service/did/resolution/resolver.go
decentralgabe: I'm fine with killing caches, but we should definitely keep storing DID Documents in storage not as a cache

decentralgabe: https://github.com/TBD54566975/ssi-service/issues/361

PR #363, opened 2023-03-29
Moarrr docs

andresuribe87:
Did a quick pass with @kirahsapong and @nitro-neal

### DIFF
--- a/pkg/server/router/credential.go
+++ b/pkg/server/router/credential.go
@@ -85,7 +85,7 @@ type CreateCredentialResponse struct {
 	// A verifiable credential conformant to the media type `application/vc+ld+json`.
 	Credential *credsdk.VerifiableCredential `json:"credential,omitempty"`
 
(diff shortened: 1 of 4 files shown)


pkg/server/router/credential.go
```diff
@@ -82,8 +82,12 @@ func (c CreateCredentialRequest) ToServiceRequest() credential.CreateCredentialR
 }
 
 type CreateCredentialResponse struct {
-	Credential    *credsdk.VerifiableCredential `json:"credential,omitempty"`
-	CredentialJWT *keyaccess.JWT                `json:"credentialJwt,omitempty"`
+	// A verifiable credential conformant to the media type `application/vc+ld+json`.
+	Credential *credsdk.VerifiableCredential `json:"credential,omitempty"`
+
+	// The same verifiable credential, but using the syntax defined for the media type `application/vc+ld+json`. See
```
decentralgabe, line 88: is this accurate? I think it should be the new media type for jwt

decentralgabe (approved):

Issue #340, opened 2023-03-24
ION Reads & Writes


decentralgabe:
Support for reading and writing ION docs.

Should uptake the work in the SSI SDK:
- https://github.com/TBD54566975/ssi-sdk/pull/307
- https://github.com/TBD54566975/ssi-sdk/pull/318

# Tasks
- [x] Implementation of the [MethodHandler Interface](https://github.com/TBD54566975/ssi-service/blob/main/pkg/service/did/handler.go#L14)
- [ ] Expansion of the API to support update, deactivate, and recover operations

decentralgabe:
Partially done in #370

Issue #362, opened 2023-03-28
Consider making the MethodHandler interface more composable


decentralgabe:
The benefits of smaller interfaces:
1. Simplicity, which means easier to understand the code.
2. Promoting reusability. Related to this, it also exposes cases for which an interface isn't needed.

For instance, it's unclear to me whether this full interface is needed. There is a lot of duplicate implementation details between `keyDIDHandler` and `webDIDHandler`. Not specific for this PR, but we should track in an issue.

_Originally posted by @andresuribe87 in https://github.com/TBD54566975/ssi-service/pull/360#discussion_r1151022354_

decentralgabe:
Consider having separate handlers for each DID method, not needing to provide <generic DID> -> <specific DID> handling logic : https://github.com/TBD54566975/ssi-service/issues/362 

cc: @nitro-neal

Issue #349, opened 2023-03-24
Support for pagination for all objects we store


decentralgabe:
Pagination support for...
- [x] DIDs
- [ ] Operations
- [x] Presentation Submissions
- [ ] Schemas
- [ ] Credentials
- [ ] Trust Establishment Docs
- [ ] ... others?

PR #370, opened 2023-04-01
ION Reads and Writes

### LINKED ISSUE #335: Generic "option" field that has unique meaning per DID method
in [pkg/server/router/did.go]

```
	KeyType crypto.KeyType `json:"keyType" validate:"required"`

	// Required when creating a DID with the `web` did method. E.g. `did:web:identity.foundation`.
	DIDWebID string `json:"didWebId"`
```
	

didWebId should this be required, BUT If we had it as required then when you would make a did-key you would have to have didWebId set to blank (or anything)

Definitely should change this, would require some refactoring on how we handle the did creation endpoint

I think we could amend this to be a generic "option" field that has unique meaning per DID method

decentralgabe:
# Overview
- Addresses part of #340 to add create and get ops for ION DIDs 
- Made DID storage more generic; switched the stored DID to an interface type and added tests for this
- Fixes #335 by adding options per DID create method
- Removed unnecessary method param for each handler
- Added a number of TODOs


pkg/server/router/keystore.go
```diff
@@ -37,14 +37,14 @@ type StoreKeyRequest struct {
...
-	// One of the following: `"Ed25519","X25519","secp256k1","P-224","P-256","P-384","P-521","RSA"`.
+	// One of the following: "Ed25519", "X25519", "secp256k1", "P-224", "P-256", "P-384", "P-521", "RSA".
 	Type crypto.KeyType `json:"type,omitempty" validate:"required"`
 
 	// See https://www.w3.org/TR/did-core/#did-controller
 	Controller string `json:"controller,omitempty" validate:"required"`
 
 	// Base58 encoding of the bytes that result from marshalling the private key using golang's implementation.
-	PrivateKeyBase58 string `json:"base58PrivateKey,omitempty" validate:"required"`
+	PrivateKeyBase58 string `json:"base58PrivateKey,omitempty"`
```
nitro-neal, line 47: remove validate:"required" was this intendend ?

pkg/service/did/ion.go
```diff
@@ -0,0 +1,289 @@
...
+	// first check if the DID is in the storage
+	gotDID := new(ionStoredDID)
+	err := h.storage.GetDID(ctx, id, gotDID)
+	if err == nil {
+		return &GetDIDResponse{DID: gotDID.DID}, nil
+	}
+	logrus.WithError(err).Warnf("error getting DID from storage: %s", id)
+
+	// if not, resolve it from the network
+	resolved, err := h.resolver.Resolve(ctx, id, nil)
```
nitro-neal, line 249: oh cool !

pkg/service/did/ion.go
```diff
@@ -0,0 +1,289 @@
...
+		Type:             keyType,
+		Controller:       controller,
+		PrivateKeyBase58: privateKeyBase58,
+	}, nil
+}
+
+func (h *ionHandler) GetDID(ctx context.Context, request GetDIDRequest) (*GetDIDResponse, error) {
+	id := request.ID
+
+	// TODO(gabe) as we are fully custodying ION DIDs this is fine; as we move to a more decentralized model we will
```
nitro-neal, line 237: maybe like we could at least have a log line / throw an error if the did in storage and the did that gets resolved are different?

pkg/service/did/ion.go
```diff
@@ -0,0 +1,289 @@
...
+
+	// if not, resolve it from the network
+	resolved, err := h.resolver.Resolve(ctx, id, nil)
+	if err != nil {
+		return nil, errors.Wrap(err, "resolving DID from network")
+	}
+	return &GetDIDResponse{DID: resolved.Document}, nil
+}
+
+// GetDIDs returns all DIDs we have in storage for ION, it is not feasible to get all DIDs from the network
```
nitro-neal, line 256: there is no Ion batch call functionality ?

pkg/service/did/storage.go
```diff
@@ -92,14 +134,27 @@ func (ds *Storage) GetDIDs(ctx context.Context, method string) ([]StoredDID, err
 		logrus.Infof("no DIDs found for method: %s", method)
 		return nil, nil
 	}
-	var stored []StoredDID
+
+	out := make([]StoredDID, 0, len(gotDIDs))
 	for _, didBytes := range gotDIDs {
-		var nextDID StoredDID
+		nextDID := reflect.New(reflect.TypeOf(outType).Elem()).Interface()
```
nitro-neal, line 140: this could be computationally really slow? (maybe not)

config/config.go
```diff
@@ -255,9 +256,9 @@ func loadDefaultServicesConfig(config *SSIServiceConfig) {
 			ServiceKeyPassword: "default-password",
 		},
 		DIDConfig: DIDServiceConfig{
-			BaseServiceConfig: &BaseServiceConfig{Name: "did"},
-			Methods:           []string{"key", "web"},
-			ResolutionMethods: []string{"key", "peer", "web", "pkh"},
+			BaseServiceConfig:      &BaseServiceConfig{Name: "did"},
+			Methods:                []string{"key", "web"},
```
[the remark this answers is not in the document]
decentralgabe: I don't think so because it requires having a valid write node configured

pkg/server/router/did.go
```diff
@@ -128,6 +133,56 @@ func (dr DIDRouter) CreateDIDByMethod(ctx context.Context, w http.ResponseWriter
 	return framework.Respond(ctx, w, resp, http.StatusCreated)
 }
 
+// toCreateDIDRequest converts CreateDIDByMethodRequest to did.CreateDIDRequest, parsing options according to method
+func toCreateDIDRequest(m didsdk.Method, request CreateDIDByMethodRequest) (*did.CreateDIDRequest, error) {
```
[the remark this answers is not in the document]
decentralgabe: help me understand...the option object can accept anything - for ion need to accept an array of public keys or service endpoints

the object is a generic `any` but to extract the value you need to know the concrete type

if it can be simplified let's do it I couldn't figure out a simpler way

pkg/server/router/keystore.go
decentralgabe: I can add this back until #371

pkg/service/did/ion.go
```diff
@@ -0,0 +1,289 @@
...
+
+	// generate the did document's initial state
+	doc := ion.Document{PublicKeys: pubKeys, Services: opts.ServiceEndpoints}
+	ionDID, createOp, err := ion.NewIONDID(doc)
+	if err != nil {
+		return nil, errors.Wrap(err, "creating new ION DID")
+	}
+
+	// submit the create operation to the ION service
+	if err = h.resolver.Anchor(ctx, createOp); err != nil {
```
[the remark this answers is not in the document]
decentralgabe: creating a DID that already exists is realistically impossible unless the anchor operation succeeds but our code does not somewhere after that...

pkg/service/did/ion.go
```diff
@@ -0,0 +1,289 @@
...
+			ServiceEndpoint: s.ServiceEndpoint,
+		})
+	}
+
+	// store the did document
+	storedDID := ionStoredDID{
+		ID:          ionDID.ID(),
+		DID:         didDoc,
+		SoftDeleted: false,
+		LongFormDID: ionDID.LongForm(),
```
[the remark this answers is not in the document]
decentralgabe: yeah I think this could end up being a few MB with updates. for now it should be fine

pkg/service/did/ion.go
decentralgabe: for now this is impossible since we only allow create and do not return the update key to the user. when we add updates and/or self-custody operations we will need to do this

pkg/service/did/ion.go
decentralgabe: looking at the api - [no](https://github.com/decentralized-identity/ion/blob/master/src/core.ts) - this is an improvement we can make

pkg/service/did/storage.go
decentralgabe: yeah, depending on how large the set of DIDs is this could be bad. this will mostly be fine with #349

config/compose.toml
```diff
@@ -38,10 +38,11 @@ password = "default-password"
 
 [services.did]
 name = "did"
-methods = ["key", "web"]
-resolution_methods = ["key", "web", "pkh", "peer"]
+methods = ["key", "web", "ion"]
+local_resolution_methods = ["key", "web", "pkh", "peer"]
 universal_resolver_url = "http://uni-resolver-web:8080"
 universal_resolver_methods = ["ion"]
+ion_resolver_url = "https://tbdwebsiteonline.com"
```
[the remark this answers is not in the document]
decentralgabe: this is in fact our ion node - you'll notice it works if you make a query like `https://tbdwebsiteonline.com/identifiers/did:ion:EiBnKzBTMxW6wM5kIlVBHidpbvKHC9QrNqfYL-Rhpbt63Q`

nitro-neal:
🏆

pkg/server/router/did.go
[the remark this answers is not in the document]
decentralgabe: oh yeah, this can work. the tradeoff is then having an "any" request vs a more generic method like I have and specific impls per type...with then the tradeoff of having some ugly mapping code.

I think the real solution is to get away from the generic -> specific funnel code we have now and instead register first-party handlers...I'm not entirely sure how this would work it would be a significant refactor but could be a more natural approach than forcing all DID methods to follow a similar structure.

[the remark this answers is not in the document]
decentralgabe: I think it's related to Andres's points here https://github.com/TBD54566975/ssi-service/issues/362 so I'll add a note with your comment and we can address in there

PR #376, opened 2023-04-10
Uptake latest SDK: Adding support for KIDs

decentralgabe:
# Overview
- Integrated key ids (kids) everywhere
- noted some future places for refactoring
- updated some error messages along the way


decentralgabe:
since a few APIs have changed to include `kid` the end to end test will be broken until after this merge. I opened https://github.com/TBD54566975/web5-components-tests/issues/17 to track

pkg/server/router/schema.go
```diff
@@ -34,8 +34,12 @@ type CreateSchemaRequest struct {
 	Author string               `json:"author" validate:"required"`
 	Name   string               `json:"name" validate:"required"`
 	Schema schemalib.JSONSchema `json:"schema" validate:"required"`
+
 	// Sign represents whether the schema should be signed by the author. Default is false.
+	// If sign is true, the schema will be signed by the author's private key with the specified KID
 	Sign bool `json:"sign"`
+	// AuthorKID represents the KID of the author's private key to sign the schema. Required if sign is true.
+	AuthorKID string `json:"authorKid"`
```
nitro-neal, line 42: why did this not work like the other APIs where its the DID and we would pull the key out of keystore to sign?

nitro-neal (commented):
Looks good! Maybe maybe make sure  anew integration test that tests the new KID features and there seems to be a broken integration test. Maybe have to generate new input data for them

nitro-neal (approved):

pkg/server/router/schema.go
decentralgabe: Since a DID can have multiple keys associated with it, we won't have a way of knowing which key to use without a KID specified.

PR #390, opened 2023-04-25
Fix ION returning DID Key objects

decentralgabe:
- fix ion returning DID key objects
- refactored DB options
- updated config examples
- now able to run the ssi service without docker again 😄

### DIFF
--- a/config/config.toml
+++ b/config/config.toml
@@ -24,24 +24,25 @@ enable_allow_all_cors = true
 
 # Storage Configuration
 [services]
-# service_endpoint = "http://localhost:3000"
+service_endpoint = "http://localhost:3000"
+
+# example bolt config with filepath option
+storage = "bolt"
+
+[[services.storage_option]]
+id = "boltdb-filepath-option"
+option = "bolt.db"
 
 # exmaple redis config with two options
-storage = "redis"
-
-[[services.storage_option]]
-id = "redis-address-option"
-option = "localhost:6379"
-
-[[services.storage_option]]
-id = "redis-password-option"
-option = "password"
-
-#storage = "bolt"
+#storage = "redis"
 #
 #[[services.storage_option]]
-#id = "boltdb-filepath-option"
-#option = "bolt.db"
+#id = "redis-address-option"
+#option = "localhost:6379"
+#
+#[[services.storage_option]]
+#id = "redis-password-option"
+#option = "password"
 
 # per-service configuration
 [services.keystore]
@@ -50,7 +51,7 @@ password = "default-password"
 
 [services.did]
 name = "did"
-methods = ["key", "web", "ion"]
+methods = ["key", "web"]
 local_resolution_methods = ["key", "web", "pkh", "peer"]
 universal_resolver_url = "http://localhost:8088"
 universal_resolver_methods = ["ion"]
--- a/config/config.toml
+++ b/config/config.toml
@@ -34,15 +34,15 @@ id = "boltdb-filepath-option"
 option = "bolt.db"
 
 # exmaple redis config with two options
-#storage = "redis"
+# storage = "redis"
 #
-#[[services.storage_option]]
-#id = "redis-address-option"
-#option = "localhost:6379"
+# [[services.storage_option]]
+# id = "redis-address-option"
+# option = "localhost:6379"
 #
-#[[services.storage_option]]
-#id = "redis-password-option"
-#option = "password"
+# [[services.storage_option]]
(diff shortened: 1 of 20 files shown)


pkg/service/did/ion.go
```diff
@@ -257,7 +257,7 @@ func (h *ionHandler) GetDID(ctx context.Context, request GetDIDRequest) (*GetDID
 func (h *ionHandler) GetDIDs(ctx context.Context) (*GetDIDsResponse, error) {
 	logrus.Debug("getting stored did:ion DIDs")
 
-	gotDIDs, err := h.storage.GetDIDs(ctx, did.KeyMethod.String(), new(ionStoredDID))
+	gotDIDs, err := h.storage.GetDIDs(ctx, did.IONMethod.String(), new(ionStoredDID))
```
decentralgabe, line 260: fix was here, rest is DB refactor

config/config.toml
```diff
@@ -25,11 +25,24 @@ enable_allow_all_cors = true
...
 
-[services.storage_option]
-address = "localhost:6379"
-password = ""
+# example bolt config with filepath option
+storage = "bolt"
+
+[[services.storage_option]]
+id = "boltdb-filepath-option"
+option = "bolt.db"
```
andresuribe87, line 34: I'm wary of running with bolt, since it might hide some issues of the default way in which we are intending to run the service.

config/config.toml
```diff
@@ -25,11 +25,24 @@ enable_allow_all_cors = true
...
+
+# exmaple redis config with two options
```
andresuribe87, line 36:
```suggestion
# example redis config with two options
```

config/config.toml
```diff
@@ -38,7 +51,7 @@ password = "default-password"
 
 [services.did]
 name = "did"
-methods = ["key", "web", "ion"]
+methods = ["key", "web"]
```
andresuribe87, line 54: Curious why this change?

pkg/storage/redis.go
```diff
@@ -60,6 +63,47 @@ func (b *RedisDB) Init(i interface{}) error {
...
+		}
+	}
+	if len(address) == 0 {
+		err = errors.New("redis address must not be empty")
+		return
+	}
+	if len(password) == 0 {
+		err = errors.New("redis password must not be empty")
+		return
+	}
```
andresuribe87, line 103: Seem repeated. Is it possible to DRY?

andresuribe87 (approved):

config/config.toml
decentralgabe: this shouldn't matter since compose/CI run with redis

decentralgabe: this shouldn't matter since compose/CI run with redis

config/config.toml
decentralgabe: having ion as a method requires a write node, and until ours is a bit more stable would like to avoid putting it here for everyone to use

pkg/storage/redis.go
decentralgabe: condensed this into a single if. need to check whether the cases were hit for required options, so think this is as simple as it can be

config/config.toml
andresuribe87: This file is the one used for local development. I don't follow why everyone would use it if we leave this in here. 

I'm also confused by why the values in this field require a write node. For instance "web" doesn't allow any writes, since there is no way or programmatically writing a did:web DID. Can you elaborate?

config/config.toml
andresuribe87: I think it matters because we're having significant differences between running así service locally, and what's being run in CI. I think we should strive to have the differences be minimal, so that it's easy to locally reproduce problems that happen in CI.

config/config.toml
decentralgabe: why does local deployment need a dependency on an ion write node?

not all DID methods require similar infra. we could refactor to separate out creating from anchoring...maybe in #340

config/config.toml
decentralgabe: there are reasons to just run the service in isolation, and to run it in a prod-like environment.

the prod-like environment is using compose. bolt is suitable for running the service in isolation.

we've also taken the approach of introducing a storage abstraction. if you're asserting that our abstraction is faulty then we should fix that -- either by making it better or removing it entirely.

decentralgabe:
@andresuribe87 going to merge, if anything blocking we can re open or discuss more in #340

PR #392, opened 2023-04-26
fix a typo in SIP list

vlad-timofeev:
# Overview
Fix a typo in the list of SIPs in the README file. Omitting the rest of the template, since it is a trivial change.


decentralgabe (approved):
Thanks for fixing!

PR #399, opened 2023-05-03
Fixing context cancelled for webhook post calls

nitro-neal:
# Overview
Fixes an issue where we would get cancelled contexts for webhook calls


pkg/service/webhook/service.go
```diff
@@ -191,9 +192,9 @@ func (s Service) PublishWebhook(ctx context.Context, noun Noun, verb Verb, paylo
 			continue
 		}
 
-		err = s.post(ctx, url, string(postJSONData))
+		err = s.post(context.Background(), url, string(postJSONData))
```
andresuribe87, line 195: `context.Background` mean that this call can wait forever. This can lead to memory increasing without a bound. 

Instead, I suggest to us some cancellable context.

config/config.toml
```diff
@@ -70,3 +70,7 @@ name = "manifest"
 
 [services.presentation]
 name = "presentation"
+
+[services.webhook]
+name = "webhook"
+webhook_timeout_in_seconds = 10
```
andresuribe87, line 76: Thoughts on using a string here and parse using the `time.ParseDuration` function? 
```suggestion
webhook_timeout = 10s
```

pkg/service/webhook/service.go
```diff
@@ -165,7 +166,7 @@ func (s Service) GetSupportedVerbs() GetSupportedVerbsResponse {
 	return GetSupportedVerbsResponse{Verbs: []Verb{Create, Delete}}
 }
 
-func (s Service) PublishWebhook(ctx context.Context, noun Noun, verb Verb, payloadReader io.Reader) {
+func (s Service) PublishWebhook(noun Noun, verb Verb, payloadReader io.Reader) {
 	webhook, err := s.storage.GetWebhook(context.Background(), string(noun), string(verb))
```
andresuribe87, line 170: This should probably use the same context, right?

config/config.toml
nitro-neal: perfect

pkg/service/webhook/service.go
nitro-neal: If I use the same context I still get:

ctx, cancel := context.WithTimeout(ctx, duration)
{"level":"warning","msg":"posting payload to http://host.docker.internal:8081/webhook with error: client http client: Post \"http://host.docker.internal:8081/webhook\": context canceled","time":"2023-05-03T18:46:57Z"}

Doing this works:
ctx, cancel := context.WithTimeout(context.background(), duration)

andresuribe87: My apologies for not being clear. I meant doing the following. So it's the same context throughout the duration of the `PublishWebhook` call. 
```go
func (s Service) PublishWebhook(noun Noun, verb Verb, payloadReader io.Reader) {
  ctx, cancel := context.WithTimeout(context.background(), duration)
  webhook, err := s.storage.GetWebhook(ctx, string(noun), string(verb))
  // some other stuff
  err = s.post(ctx, url, string(postJSONData))
  // .. even more stuff
}
```

config/config.go
```diff
@@ -283,6 +284,7 @@ func loadDefaultServicesConfig(config *SSIServiceConfig) {
 		},
 		WebhookConfig: WebhookServiceConfig{
 			BaseServiceConfig: &BaseServiceConfig{Name: "webhook"},
+			WebhookTimeout:    "10s",
```
decentralgabe, line 287: should this be a `time.Duration` value in go parsed from a string in config?

pkg/service/webhook/service.go
```diff
@@ -191,10 +193,18 @@ func (s Service) PublishWebhook(ctx context.Context, noun Noun, verb Verb, paylo
 			continue
 		}
 
+		duration, err := time.ParseDuration(s.config.WebhookTimeout)
+		if err != nil {
+			logrus.Error("problem parsing webhook timeout")
```
decentralgabe, line 198:
```suggestion
			logrus.WithError(err).Error("problem parsing webhook timeout")
```

pkg/service/webhook/service.go
```diff
@@ -191,10 +193,18 @@ func (s Service) PublishWebhook(ctx context.Context, noun Noun, verb Verb, paylo
...
+			return
+		}
+		ctx, cancel := context.WithTimeout(context.Background(), duration)
+		defer cancel()
 		err = s.post(ctx, url, string(postJSONData))
 		if err != nil {
-			logrus.Warnf("posting payload to %s", url)
+			logrus.Warnf("posting payload to %s with error: %s", url, err)
```
decentralgabe, line 205:
```suggestion
			logrus.WithError(err).Warnf("posting payload to %s", url)
```

pkg/service/webhook/service.go
```diff
@@ -191,10 +193,18 @@ func (s Service) PublishWebhook(ctx context.Context, noun Noun, verb Verb, paylo
...
 		}
 
+		duration, err := time.ParseDuration(s.config.WebhookTimeout)
+		if err != nil {
+			logrus.Error("problem parsing webhook timeout")
+			return
+		}
+		ctx, cancel := context.WithTimeout(context.Background(), duration)
+		defer cancel()
 		err = s.post(ctx, url, string(postJSONData))
```
decentralgabe, line 203: can combine with the line below

pkg/service/webhook/service.go
nitro-neal: Good call !

config/config.go
nitro-neal: I think keeping it a string would be more consistent to the rest of the code and not lock you into using one duration framework, also because of the shortcomings of toml I would have to make a messy wrapper and custom logic to parse a string to a time.Duration object.

pkg/service/webhook/service.go
nitro-neal: This has changed locatoin

pkg/service/webhook/service.go
```diff
@@ -165,8 +166,17 @@ func (s Service) GetSupportedVerbs() GetSupportedVerbsResponse {
 	return GetSupportedVerbsResponse{Verbs: []Verb{Create, Delete}}
 }
 
-func (s Service) PublishWebhook(ctx context.Context, noun Noun, verb Verb, payloadReader io.Reader) {
-	webhook, err := s.storage.GetWebhook(context.Background(), string(noun), string(verb))
+func (s Service) PublishWebhook(noun Noun, verb Verb, payloadReader io.Reader) {
+	duration, err := time.ParseDuration(s.config.WebhookTimeout)
```
andresuribe87, line 170: nit: parse the duration when creating the service, instead of doing it for every call.

pkg/service/webhook/service.go
```diff
@@ -165,8 +166,17 @@ func (s Service) GetSupportedVerbs() GetSupportedVerbsResponse {
...
+	if err != nil {
+		logrus.WithError(err).Error("problem parsing webhook timeout")
```
andresuribe87, line 172:
```suggestion
		logrus.WithError(err).Error("parsing webhook timeout")
```

andresuribe87 (approved):
Awesome! Couple of small nits.

PR #402, opened 2023-05-04
Update GetDIDs to accept a deleted query parameter

kirahsapong:
# Overview
This PR introduces a `deleted` query parameter to the GetDIDs API call. 
**WIP - some sections not completed yet, more work to be done to this PR**

## Description
When calling GetDIDs, the presence of a query param `deleted="true"` should return only DIDs that have been soft deleted (has `isSoftDeleted` flag), while calling GetDIDs without the presence of the query param should return only DIDs that have not been soft deleted.  

## How Has This Been Tested?
- [x] Functionally tested in Postman 
1. Ran through steps of creating did, "soft" deleting did, getting all dids, getting all dids with query param `deleted="true"`
2. Confirmed getDIDs returned empty, getDIDs with `deleted="true"` returned the "soft" deleted did
- [x] Added unit tests where `deleted="true"` should return all deleted dids
1. Run `mage test`
2. All tests should pass

## TODO
- Add Swagger Documentation for new optional `deleted` query param

### DIFF
--- a/pkg/server/router/did.go
+++ b/pkg/server/router/did.go
@@ -3,7 +3,6 @@ package router
 import (
 	"context"
 	"fmt"
-	"log"
 	"net/http"
 	"strconv"
 
@@ -246,7 +245,7 @@ type GetDIDsRequest struct {
 // @Success     200     {object} GetDIDsByMethodResponse
 // @Failure     400     {string} string "Bad request"
 // @Failure     500     {string} string "Internal server error"
-// @Router      /v1/dids [get]
+// @Router      /v1/dids/{method} [get]
 func (dr DIDRouter) GetDIDsByMethod(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
 	method := framework.GetParam(ctx, MethodParam)
 	deleted := framework.GetQueryValue(r, DeletedParam)
@@ -255,19 +254,21 @@ func (dr DIDRouter) GetDIDsByMethod(ctx context.Context, w http.ResponseWriter, 
 		logrus.Error(errMsg)
 		return framework.NewRequestErrorMsg(errMsg, http.StatusBadRequest)
 	}
-	isDeleted := false
+	getIsDeleted := false
 	if deleted != nil {
 		checkDeleted, err := strconv.ParseBool(*deleted)
-		isDeleted = checkDeleted
+		getIsDeleted = checkDeleted
 
 		if err != nil {
-			log.Fatal(err)
+			errMsg := "get DIDs by method request encountered a problem with the `deleted` query param"
+			logrus.Error(errMsg)
+			return framework.NewRequestErrorMsg(errMsg, http.StatusBadRequest)
 		}
 	}
 
 	// TODO(gabe) check if the method is supported, to tell whether this is a bad req or internal error
 	// TODO(gabe) differentiate between internal errors and not found DIDs
-	getDIDsRequest := did.GetDIDsRequest{Method: didsdk.Method(*method), Deleted: isDeleted}
+	getDIDsRequest := did.GetDIDsRequest{Method: didsdk.Method(*method), Deleted: getIsDeleted}
 	gotDIDs, err := dr.service.GetDIDsByMethod(ctx, getDIDsRequest)
 	if err != nil {
 		errMsg := fmt.Sprintf("could not get DIDs for method: %s", *method)
--- a/pkg/server/router/did.go
+++ b/pkg/server/router/did.go
@@ -241,7 +241,7 @@ type GetDIDsRequest struct {
 // @Tags        DecentralizedIdentityAPI
 // @Accept      json
 // @Produce     json
-// @Param       deleted  query    boolean false
+// @Param       deleted  query    boolean false "When true, returns soft-deleted DIDs. Otherwise, returns DIDs that have not been soft-deleted. Default is false."
 // @Param       request body     GetDIDsRequest true "request body"
 // @Success     200     {object} GetDIDsByMethodResponse
 // @Failure     400     {string} string "Bad request"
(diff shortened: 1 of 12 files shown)


pkg/server/router/did.go
```diff
@@ -250,17 +253,24 @@ type GetDIDsRequest struct {
...
 	if method == nil {
 		errMsg := "get DIDs by method request missing method parameter"
 		logrus.Error(errMsg)
 		return framework.NewRequestErrorMsg(errMsg, http.StatusBadRequest)
 	}
 
+	isDeleted, err := strconv.ParseBool(*deleted)
+
+	if err != nil {
+		log.Fatal(err)
```
nitro-neal, line 268: logrus.Error(errMsg)
		return framework.NewRequestErrorMsg("cannot parse deleted query value", http.StatusBadRequest)

pkg/service/did/handler.go
```diff
@@ -16,7 +16,7 @@ type MethodHandler interface {
 	CreateDID(ctx context.Context, request CreateDIDRequest) (*CreateDIDResponse, error)
 	// TODO(gabe): support query parameters to get soft deleted and other DIDs https://github.com/TBD54566975/ssi-service/issues/364
 	GetDID(ctx context.Context, request GetDIDRequest) (*GetDIDResponse, error)
-	GetDIDs(ctx context.Context) (*GetDIDsResponse, error)
+	GetDIDs(ctx context.Context, deleted bool) (*GetDIDsResponse, error)
```
decentralgabe, line 19: can you encapsulate this into a `GetDIDsRequest`?

pkg/server/router/did.go
```diff
@@ -250,17 +253,24 @@ type GetDIDsRequest struct {
...
+func (dr DIDRouter) GetDIDsByMethod(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
 	method := framework.GetParam(ctx, MethodParam)
+	deleted := framework.GetQueryValue(r, DeletedParam)
 	if method == nil {
 		errMsg := "get DIDs by method request missing method parameter"
 		logrus.Error(errMsg)
 		return framework.NewRequestErrorMsg(errMsg, http.StatusBadRequest)
 	}
 
+	isDeleted, err := strconv.ParseBool(*deleted)
```
nitro-neal, line 265: probably need to handle the case where *deleted is nil (if someone does a getDids request without a query param)

pkg/service/did/ion.go
```diff
@@ -263,9 +263,10 @@ func (h *ionHandler) GetDIDs(ctx context.Context) (*GetDIDsResponse, error) {
 	}
 	dids := make([]did.Document, 0, len(gotDIDs))
 	for _, gotDID := range gotDIDs {
-		if !gotDID.IsSoftDeleted() {
+		if (deleted && gotDID.IsSoftDeleted()) || (!deleted && !gotDID.IsSoftDeleted()) {
```
nitro-neal, line 266: need to add this to key, web

pkg/service/did/ion.go
```diff
@@ -287,3 +288,19 @@ func (h *ionHandler) SoftDeleteDID(ctx context.Context, request DeleteDIDRequest
 
 	return h.storage.StoreDID(ctx, *gotDID)
 }
+
+func (h *ionHandler) GetDeletedDIDs(ctx context.Context) (*GetDIDsResponse, error) {
```
nitro-neal, line 292: dont think we are using this anymore

nitro-neal (commented):
Will need to do linting, fix build, maybe add a unit test

pkg/service/did/ion.go
decentralgabe: we should just move this logic to the DB I know @andresuribe87 has mentioned this before

pkg/server/router/did.go
```diff
@@ -250,17 +253,27 @@ type GetDIDsRequest struct {
...
 		logrus.Error(errMsg)
 		return framework.NewRequestErrorMsg(errMsg, http.StatusBadRequest)
 	}
+	isDeleted := false
+	if deleted != nil {
+		checkDeleted, err := strconv.ParseBool(*deleted)
+		isDeleted = checkDeleted
+
+		if err != nil {
+			log.Fatal(err)
```
decentralgabe, line 270: don't panic! change this to what you have above

```
logrus.Error(errMsg)
return framework.NewRequestErrorMsg(errMsg, http.StatusBadRequest)
```

pkg/service/did/ion.go
```diff
@@ -254,18 +254,16 @@ func (h *ionHandler) GetDID(ctx context.Context, request GetDIDRequest) (*GetDID
 }
 
 // GetDIDs returns all DIDs we have in storage for ION, it is not feasible to get all DIDs from the network
-func (h *ionHandler) GetDIDs(ctx context.Context) (*GetDIDsResponse, error) {
+func (h *ionHandler) GetDIDs(ctx context.Context, deleted bool) (*GetDIDsResponse, error) {
```
decentralgabe, line 257: consider a more meaningful variable name than `deleted`. it is ambiguous whether the presence means you're only getting deleted DIDs, or you're including them. I believe you want the including behavior. in this case `includeDeleted` is a more meaningful name

pkg/service/did/storage_test.go
```diff
@@ -202,7 +202,7 @@ func TestStorage(t *testing.T) {
 		assert.Contains(tt, err.Error(), "could not get DID: did:key:test-1")
 
 		// get both and verify there is one
-		gotDIDs, err = ds.GetDIDsDefault(context.Background(), didsdk.KeyMethod.String())
+		gotDIDs, err = ds.GetDIDsDefault(context.Background(), didsdk.KeyMethod.String(), false)
```
decentralgabe, line 205: should add a case here to test where deleted = true?

pkg/service/did/ion.go
decentralgabe: it may be cleaner to keep the code as it was internally. meaning we have a separate `GetDeletedDIDs` method and just the API can interpret the flag.

do need to support all three cases?
- get all DIDs 
- get all DIDs and deleted DIDs
- get just deleted DIDs

I think we should

decentralgabe: sure happy to support those two

decentralgabe (approved):
lgtm

nitro-neal (approved):
Looks good!

PR #409, opened 2023-05-08
update go

nitro-neal:
Bumping go version


decentralgabe:
@nitro-neal can you do a cmd+f and find all references to go 1.20.3 and change them to 1.20.4?
there should be some in the README, CONTRIBUTING, ci files, etc

PR #420, opened 2023-05-10
Support external KMS for service keys.

andresuribe87:
# Overview
Add support so Google Cloud KMS and AWS KMS can be used to store the Encrypt and Decrypt all stored keys in SSI service using the KMS managed Key Encryption Key. 

# Description
This change refactors the KeyStorageService so it can use any struct that implements Encryptor and Decryptor. For the implementation that's backed by KMS, I added tink. (learn more about it [here](https://developers.google.com/tink)).

Operators of the service can choose which to use by using the config values exposed.

# How Has This Been Tested?
All unit tests pass. Beyond that, it's very tough to test this integration without an actual working KMS, as this is doing a configuration change.


config/config.go
```diff
@@ -98,7 +98,15 @@ type KeyStoreServiceConfig struct {
 	*BaseServiceConfig
 	// Service key password. Used by a KDF whose key is used by a symmetric cypher for key encryption.
 	// The password is salted before usage.
+	// Note that this field is only used when MasterKeyURI is empty.
 	ServiceKeyPassword string `toml:"password"`
+
+	// The URI for the master key. We use tink for envelope encryption as described in https://github.com/google/tink/blob/9bc2667963e20eb42611b7581e570f0dddf65a2b/docs/KEY-MANAGEMENT.md#key-management-with-tink
+	// When left empty, then ServiceKeyPassword is used.
+	MasterKeyURI string `toml:"master_key_uri"`
```
decentralgabe, line 106: `ServiceKeyURI`?

pkg/service/keystore/storage.go
```diff
@@ -48,43 +56,107 @@ const (
...
-	serviceKey []byte
+	db        storage.ServiceStorage
+	encryptor Encryptor
+	decryptor Decryptor
 }
 
-func NewKeyStoreStorage(db storage.ServiceStorage, key ServiceKey) (*Storage, error) {
-	keyBytes, err := base58.Decode(key.Base58Key)
+func NewKeyStoreStorage(db storage.ServiceStorage, e Encryptor, d Decryptor) (*Storage, error) {
+	bolt := &Storage{
```
decentralgabe, line 65: `storage` instead of `bolt`?

pkg/service/keystore/storage.go
```diff
@@ -48,43 +56,107 @@ const (
...
+	}
+
+	return bolt, nil
+}
+
+type wrappedEncryptor struct {
+	tink.AEAD
+}
+
+func (w wrappedEncryptor) Encrypt(_ context.Context, plaintext, contextData []byte) ([]byte, error) {
```
decentralgabe, line 78: why take in a context?

pkg/service/keystore/storage.go
```diff
@@ -48,43 +56,107 @@ const (
...
+		Base58Salt: serviceKeySalt,
 	}
-	return bolt, nil
+	if err := storeServiceKey(ctx, db, key); err != nil {
+		return nil, nil, err
+	}
+	return &encryptor{db}, &decryptor{db}, nil
+}
+
+func NewExternalEncryptor(ctx context.Context, cfg config.KeyStoreServiceConfig) (Encryptor, Decryptor, error) {
```
decentralgabe, line 119: nit: here and elsewhere `encryptor` and `decryptor` are british spellings, prefer `encrypter` and `decrypter`

config/config.go
andresuribe87: I'd like to keep the same terminology as in https://github.com/google/tink/blob/9bc2667963e20eb42611b7581e570f0dddf65a2b/docs/KEY-MANAGEMENT.md#key-management-systems

pkg/service/keystore/storage.go
```diff
@@ -48,43 +56,107 @@ const (
...
+const (
+	gcpKMSScheme = "gcp-kms"
+	awsKMSScheme = "aws-kms"
+)
+
+func NewEncryption(db storage.ServiceStorage, cfg config.KeyStoreServiceConfig) (Encryptor, Decryptor, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	if strings.HasPrefix(cfg.MasterKeyURI, gcpKMSScheme) || strings.HasPrefix(cfg.MasterKeyURI, awsKMSScheme) {
```
decentralgabe, line 99: should we log a warning if we get an unknown scheme?

pkg/service/keystore/storage.go
```diff
@@ -48,43 +56,107 @@ const (
...
+
+type wrappedDecryptor struct {
+	tink.AEAD
+}
+
+func (w wrappedDecryptor) Decrypt(_ context.Context, ciphertext, contextInfo []byte) ([]byte, error) {
+	return w.AEAD.Decrypt(ciphertext, contextInfo)
+}
+
+const (
```
decentralgabe, line 90: where are these prefixes specified? how will someone configuring the service know to use them?

pkg/service/keystore/storage.go
andresuribe87: Because `wrappedEncryptor` needs to implement the `Encryptor` interface

pkg/service/keystore/storage.go
```diff
@@ -48,43 +56,107 @@ const (
...
+		client, err = gcpkms.NewClientWithOptions(ctx, cfg.MasterKeyURI, option.WithCredentialsFile(cfg.KMSCredentialsPath))
+		if err != nil {
+			return nil, nil, errors.Wrap(err, "creating gcp kms client")
+		}
+	} else if strings.HasPrefix(cfg.MasterKeyURI, awsKMSScheme) {
+		client, err = awskms.NewClientWithCredentials(cfg.MasterKeyURI, cfg.KMSCredentialsPath)
+		if err != nil {
+			return nil, nil, errors.Wrap(err, "creating aws kms client")
+		}
+	}
```
decentralgabe, line 132: what happens with the else case?

pkg/service/keystore/storage.go
```diff
@@ -48,43 +56,107 @@ const (
...
-	keyBytes, err := base58.Decode(key.Base58Key)
+func NewKeyStoreStorage(db storage.ServiceStorage, e Encryptor, d Decryptor) (*Storage, error) {
+	bolt := &Storage{
+		db:        db,
+		encryptor: e,
+		decryptor: d,
+	}
+
+	return bolt, nil
+}
```
andresuribe87, line 72:
```suggestion
	storage := &Storage{
		db:        db,
		encryptor: e,
		decryptor: d,
	}

	return storage, nil
}
```

pkg/service/keystore/storage.go
```diff
@@ -102,10 +174,59 @@ func (kss *Storage) getAndSetServiceKey(ctx context.Context) ([]byte, error) {
...
+
+func (e encryptor) Encrypt(ctx context.Context, plaintext, _ []byte) ([]byte, error) {
+	// get service key
+	serviceKey, err := getServiceKey(ctx, e.db)
+	if err != nil {
+		return nil, err
+	}
+	// encrypt key before storing
+	encryptedKey, err := util.XChaCha20Poly1305Encrypt(serviceKey, plaintext)
```
decentralgabe, line 203: can we always use `XChaCha20Poly1305`? or is this just if we generate a key ourselves?
since if a KMS is configured it could use any of these keys https://developers.google.com/tink/aead

pkg/service/keystore/storage.go
```diff
@@ -48,43 +56,107 @@ const (
...
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "creating aead from key handl")
+	}
+	return wrappedEncryptor{a}, wrappedDecryptor{a}, nil
 }
 
 // TODO(gabe): support more robust service key operations, including rotation, and caching
-func (kss *Storage) storeServiceKey(ctx context.Context, key ServiceKey) error {
+func storeServiceKey(ctx context.Context, db storage.ServiceStorage, key ServiceKey) error {
 	keyBytes, err := json.Marshal(key)
```
decentralgabe, line 148: can check if the key is empty

pkg/service/keystore/storage.go
decentralgabe: nvm I see this is only for our service key

for the external key - is there no service key? does this fail?

config/config.go
decentralgabe: then let's rename `ServiceKeyPassword` to `MasterKeyPassword` for consistency

pkg/service/keystore/storage.go
andresuribe87: Switched this to return an error.

pkg/service/keystore/storage.go
andresuribe87: I'd rather leave as it, as this was refactoring. I don't want to change behaviors if possible.

pkg/service/keystore/storage.go
andresuribe87: If there is no service key, this never gets called. Details below. 

The struct `encrypter` is used only when `cfg.MasterKeyURI` is empty. It represents the old way in which we were doing things.

When using an external KMS, then the struct that's used is `wrappedEncrypter`. It delegates signing to the external KMS.

config/config.go
andresuribe87: Done.

pkg/service/keystore/storage.go
decentralgabe: it would be good to add a brief doc in `docs` about using a KMS or perhaps the config docs

pkg/service/keystore/storage.go
```diff
@@ -48,43 +56,110 @@ const (
...
+	case strings.HasPrefix(cfg.MasterKeyURI, awsKMSScheme):
+		client, err = awskms.NewClientWithCredentials(cfg.MasterKeyURI, cfg.KMSCredentialsPath)
+		if err != nil {
+			return nil, nil, errors.Wrap(err, "creating aws kms client")
+		}
+	default:
+		return nil, nil, errors.Errorf("master_key_uri value %q is not supported", cfg.MasterKeyURI)
+	}
+	registry.RegisterKMSClient(client)
+	dek := aead.XChaCha20Poly1305KeyTemplate()
```
decentralgabe, line 137: so we are assuming they have a XChaCha20Poly1305 key, if we're forcing this (I think we should for now at least) then this needs to be documented on how they configure their KMS

decentralgabe (approved):
awesome

PR #431, opened 2023-05-16
Detect if swagger is out of date; publish on merge

decentralgabe:
- on each PR fail a check if swagger is out of date
- on merge of the PR trigger the netlify webhook


michaelneale:
sounds good - so it up to developer making a change to update the swagger file? (this is probably a good thing as it means an api change, which everyone needs to be aware of)

decentralgabe:
correct. you're told if you break it and you fix it. the action keeps the main branch up to date.

decentralgabe:
seems this is failing even after running mage spec locally. I see the old comment here which found the same issue https://github.com/TBD54566975/ssi-service/pull/198#issuecomment-1378289552

the real solution is to just have CI generate it, I think

michaelneale:
@decentralgabe hrm this sounds familiar, like it wasn't entirely deterministic so it couldn't know if it was generated correctly or not (which may have been why we yeeted it to another branch in the first place). A distant memory. Annoying.

michaelneale:
perhaps there is a way to do a semantic diff somehow ...

andresuribe87:
Yeah not really sure what was broken here that needed fixing? Is the fact that we're pinning to a specific version of swagger a problem?

> the real solution is to just have CI generate it, I think, or not rely on a static file

CI does generate the file right now. It's written to the `main_swagger` branch. That's what gets published.

andresuribe87:
FWIW, the swagger version is important to pin. If it isn't pinned, then the documentation generation may fail. It's something to do with the recursive deps option of the library not playing nice with a library we import.

decentralgabe:
> Yeah not really sure what was broken here that needed fixing? Is the fact that we're pinning to a specific version of swagger a problem?

Having it on a branch is broken IMO, since people who check out main and look at swagger get an outdated view

PR #432, opened 2023-05-16
Rename presentation definition to presentation request

### LINKED ISSUE #375: Rename presentation definition endpoints
Presentation Requests are signed versions of Presentation Definitions, intended for specific audiences.
Today, our API for Presentation Definition returns signed objects when it should not, as Presentation Definitions themselves aren't signed.

andresuribe87:
# Overview
Fixes #375 by renaming endpoints and internal models so it's clear that we are storing presentation requests rather than definitions.

# Description
This is a refactor. While it's big, it's still simple. 

# How Has This Been Tested?
Unit and integration tests still pass.


integration/common.go
```diff
@@ -217,18 +217,18 @@ func CreateCredentialApplicationJWT(credApplication credApplicationParams, crede
...
-type definitionParams struct {
+type presentationRequestParams struct {
 	Author    string
 	AuthorKID string
 }
 
-func CreatePresentationDefinition(params definitionParams) (string, error) {
-	logrus.Println("\n\nCreate our Presentation Definition:")
-	definitionJSON, err := resolveTemplate(params, "presentation-definition-input.json")
+func CreatePresentationRequest(params presentationRequestParams) (string, error) {
```
decentralgabe, line 225: I'm not sure about this. I see a value in storing a presentation definition, and then generating a presentation request separately, namely because they might have different wrapping containers.

I would be open to changing the endpoints to `CreatePresentationDefintion` and `GetPresentationRequest` but I'd really rather see this PR revert the Presentation Definition changes and add a new `GetPresentationRequest` endpoint which returns a signed Presentation Definition.

decentralgabe (changes requested):

pkg/server/server.go
```diff
@@ -21,25 +21,25 @@ import (
...
+	ReadinessPrefix            = "/readiness"
+	V1Prefix                   = "/v1"
+	OperationPrefix            = "/operations"
+	DIDsPrefix                 = "/dids"
+	ResolverPrefix             = "/resolver"
+	SchemasPrefix              = "/schemas"
+	CredentialsPrefix          = "/credentials"
+	StatusPrefix               = "/status"
+	PresentationsPrefix        = "/presentations"
+	PresentationRequestsPrefix = "/requests"
```
decentralgabe, line 34: definition still is a top level construct we want to maintain

integration/common.go
decentralgabe: > What I'm doing here is just naming things correctly, which I believe is important to do. Adding a new endpoint wouldn't fix the incorrect names.

Creating a Presentation Definition doesn't take in a signed request. It should store the Presentation Definition object.

> Can you elaborate the value of storing a presentation definition separate from the request in which it's going to be transported in?

Yes. Presentations are often created for specific parties. Meaning they have an audience (`aud` in jwt land), and expiration (`exp`) to denote how long they're valid for. Sometimes we can create longer-lived presentations, and presentations that don't have a specific audience. This design boxes us into that last case.

PR #442, opened 2023-05-18
gin swagger; remove streamlit

### LINKED ISSUE #438: [Server] Use gin for serving swagger docs
https://github.com/swaggo/gin-swagger

decentralgabe:
fixes #438 

Cannot merge this until we figure out how to get a build user to push to main with updated swagger docs. See: https://github.com/orgs/community/discussions/25305


cmd/ssiservice/main.go
```diff
@@ -53,6 +52,12 @@ func run() error {
 		logrus.Fatalf("could not instantiate config: %s", err.Error())
 	}
 
+	// set up some additional swagger config
+	doc.SwaggerInfo.Version = cfg.SVN
```
decentralgabe, line 56: adding the `docs.go` file is necessary to do this config

.github/workflows/ci.yml
```diff
@@ -49,8 +51,14 @@ jobs:
     - name: Test
       run: mage CITest
 
-    - name: Generate Swagger
-      run: mage spec
+    - name: Update spec and push generated spec file
+      if: github.event_name == 'pull_request' && github.event.action != 'closed' && github.event.pull_request.merged != true
+      run: |
+        mage spec
```
andresuribe87, line 57: `mage spec` is also doing formatting of files. See https://github.com/TBD54566975/ssi-service/blob/7047be55d3d57ca5e4ee4919126e4acdeb275444/magefile.go#L114-L117

I think that needs to be broken up into separate mage targets. 

As it stands, you might get a passing CI during branch review, but failing in main because of this (a linter might fail).

andresuribe87 (approved):
One minor nit to make this more robust.

.github/workflows/ci.yml
decentralgabe: done

PR #443, opened 2023-05-19
Factored request part out of the definition endpoint

andresuribe87:
# Overview
This addresses #375 by creating a separate endpoint for it.

# Description
This follows the same pattern as definitions/submissions in the `presentation` package.

# How Has This Been Tested?
Unit tests.

### DIFF
--- a/pkg/server/router/presentation.go
+++ b/pkg/server/router/presentation.go
@@ -3,9 +3,11 @@ package router
 import (
 	"fmt"
 	"net/http"
+	"time"
 
 	"github.com/TBD54566975/ssi-sdk/credential"
 	"github.com/TBD54566975/ssi-sdk/credential/exchange"
+	sdkutil "github.com/TBD54566975/ssi-sdk/util"
 	"github.com/gin-gonic/gin"
 	"github.com/goccy/go-json"
 	"github.com/pkg/errors"
@@ -40,13 +42,6 @@ type CreatePresentationDefinitionRequest struct {
 	Format                 *exchange.ClaimFormat            `json:"format,omitempty" validate:"omitempty,dive"`
 	InputDescriptors       []exchange.InputDescriptor       `json:"inputDescriptors" validate:"required,dive"`
 	SubmissionRequirements []exchange.SubmissionRequirement `json:"submissionRequirements,omitempty" validate:"omitempty,dive"`
-
-	// DID of the author of this presentation definition. The DID must have been previously created with the DID API,
-	// or the PrivateKey must have been added independently.
-	Author string `json:"author" validate:"required"`
-	// The privateKey associated with the KID will be used to sign an envelope that contains
-	// the created presentation definition.
-	AuthorKID string `json:"authorKid" validate:"required"`
 }
 
 type CreatePresentationDefinitionResponse struct {
@@ -86,16 +81,13 @@ func (pr PresentationRouter) CreateDefinition(c *gin.Context) error {
 	}
 	serviceResp, err := pr.service.CreatePresentationDefinition(c, model.CreatePresentationDefinitionRequest{
 		PresentationDefinition: *def,
-		Author:                 request.Author,
-		AuthorKID:              request.AuthorKID,
 	})
 	if err != nil {
 		return framework.LoggingRespondErrWithMsg(c, err, errMsg, http.StatusInternalServerError)
 	}
 
 	resp := CreatePresentationDefinitionResponse{
-		PresentationDefinition:    serviceResp.PresentationDefinition,
-		PresentationDefinitionJWT: serviceResp.PresentationDefinitionJWT,
+		PresentationDefinition: serviceResp.PresentationDefinition,
 	}
 	return framework.Respond(c, resp, http.StatusCreated)
 }
@@ -133,10 +125,6 @@ func definitionFromRequest(request CreatePresentationDefinitionRequest) (*exchan
 
 type GetPresentationDefinitionResponse struct {
 	PresentationDefinition exchange.PresentationDefinition `json:"presentation_definition,omitempty"`
-
-	// Signed envelope that contains the PresentationDefinition created using the privateKey of the author of the
-	// definition.
-	PresentationDefinitionJWT keyaccess.JWT `json:"presentationDefinitionJWT,omitempty"`
 }
 
 // GetDefinition godoc
@@ -164,8 +152,7 @@ func (pr PresentationRouter) GetDefinition(c *gin.Context) error {
 	}
 
 	resp := GetPresentationDefinitionResponse{
-		PresentationDefinition:    def.PresentationDefinition,
-		PresentationDefinitionJWT: def.PresentationDefinitionJWT,
+		PresentationDefinition: def.PresentationDefinition,
 	}
 	return framework.Respond(c, resp, http.StatusOK)
 }
@@ -227,6 +214,9 @@ func (pr PresentationRouter) DeleteDefinition(c *gin.Context) error {
 }
 
 type CreateSubmissionRequest struct {
+	// A Verifiable Presentation that's encoded as a JWT.
+	// Verifiable Presentation are described in https://www.w3.org/TR/vc-data-model/#presentations-0
+	// JWT encoding of the Presentation as described in https://www.w3.org/TR/vc-data-model/#presentations-0
 	SubmissionJWT keyaccess.JWT `json:"submissionJwt" validate:"required"`
 }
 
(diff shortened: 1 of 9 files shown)


pkg/server/router/presentation.go
```diff
@@ -40,13 +42,6 @@ type CreatePresentationDefinitionRequest struct {
 	Format                 *exchange.ClaimFormat            `json:"format,omitempty" validate:"omitempty,dive"`
 	InputDescriptors       []exchange.InputDescriptor       `json:"inputDescriptors" validate:"required,dive"`
 	SubmissionRequirements []exchange.SubmissionRequirement `json:"submissionRequirements,omitempty" validate:"omitempty,dive"`
```
decentralgabe, line 44: I think having author is still valuable without kid

pkg/server/router/presentation.go
```diff
@@ -444,3 +434,143 @@ func (pr PresentationRouter) ReviewSubmission(c *gin.Context) error {
...
+const DefaultExpirationDuration = 30 * time.Minute
+
+type CreateRequestRequest struct {
+	// Audience as defined in https://www.rfc-editor.org/rfc/rfc7519.html#section-4.1.3
+	// Optional
+	Audience []string `json:"audience"`
+
+	// Expiration as defined in https://www.rfc-editor.org/rfc/rfc7519.html#section-4.1.4
+	// Optional. When not specified, the request will be valid for 30 minutes.
+	Expiration string `json:"expiration"`
```
decentralgabe, line 447: hmm...what do you think about turning this off by default? I feel like no expiry is a more common behavior

pkg/server/router/presentation.go
```diff
@@ -444,3 +434,143 @@ func (pr PresentationRouter) ReviewSubmission(c *gin.Context) error {
...
+
+	// DID of the issuer of this presentation definition. The DID must have been previously created with the DID API,
+	// or the PrivateKey must have been added independently.
+	IssuerDID string `json:"issuerID" validate:"required"`
```
decentralgabe, line 451:
```suggestion
	IssuerDID string `json:"issuerId" validate:"required"`
```

pkg/server/router/presentation.go
```diff
@@ -444,3 +434,143 @@ func (pr PresentationRouter) ReviewSubmission(c *gin.Context) error {
...
+
+	// The privateKey associated with the KID will be used to sign an envelope that contains
+	// the created presentation definition.
+	KID string `json:"kid" validate:"required"`
```
decentralgabe, line 455:
```suggestion
	IssuerKID string `json:"issuerKid" validate:"required"`
```

for consistency

pkg/service/presentation/model/model.go
```diff
@@ -124,3 +121,50 @@ func ServiceModel(storedSubmission *storage.StoredSubmission) Submission {
 		VerifiablePresentation: &storedSubmission.VerifiablePresentation,
 	}
 }
+
+type CreateRequestRequest struct {
+	PresentationRequest Request `json:"presentationRequest"`
+}
+
+type CreateRequestResponse struct {
+	PresentationRequest Request `json:"presentationRequest"`
```
decentralgabe, line 130: do we need to return it back, or is a 201 ok?

pkg/service/presentation/model/model.go
```diff
@@ -124,3 +121,50 @@ func ServiceModel(storedSubmission *storage.StoredSubmission) Submission {
...
+	ID string `json:"id" validate:"required"`
+}
+
+type Request struct {
+	// ID for this request. It matches the "jti" claim in the JWT.
+	// This is an output only field.
+	ID string `json:"id"`
+
+	// Audience as defined in https://www.rfc-editor.org/rfc/rfc7519.html#section-4.1.3.
+	Audience []string `json:"audience"`
```
decentralgabe, line 152:
```suggestion
	Audience []string `json:"audience,omitempty"`
```

pkg/service/presentation/model/model.go
```diff
@@ -124,3 +121,50 @@ func ServiceModel(storedSubmission *storage.StoredSubmission) Submission {
...
+}
+
+type DeleteRequestRequest struct {
+	ID string `json:"id" validate:"required"`
+}
+
+type Request struct {
+	// ID for this request. It matches the "jti" claim in the JWT.
+	// This is an output only field.
+	ID string `json:"id"`
```
decentralgabe, line 149:
```suggestion
	ID string `json:"id,omitempty"`
```

pkg/service/presentation/model/model.go
```diff
@@ -124,3 +121,50 @@ func ServiceModel(storedSubmission *storage.StoredSubmission) Submission {
...
+
+	// Audience as defined in https://www.rfc-editor.org/rfc/rfc7519.html#section-4.1.3.
+	Audience []string `json:"audience"`
+
+	// Expiration as defined in https://www.rfc-editor.org/rfc/rfc7519.html#section-4.1.4
+	Expiration time.Time `json:"expiration" validate:"required"`
+
+	// DID of the issuer of this presentation definition.
+	IssuerDID string `json:"issuerID" validate:"required"`
```
decentralgabe, line 158:
```suggestion
	IssuerDID string `json:"issuerId" validate:"required"`
```

pkg/service/presentation/model/model.go
```diff
@@ -124,3 +121,50 @@ func ServiceModel(storedSubmission *storage.StoredSubmission) Submission {
...
+
+	// The privateKey associated with the KID used to sign the JWT.
+	KID string `json:"kid" validate:"required"`
```
decentralgabe, line 161:
```suggestion
	IssuerKID string `json:"issuerKid" validate:"required"`
```

pkg/service/presentation/service.go
```diff
@@ -27,7 +30,7 @@ import (
 )
 
 type Service struct {
-	storage    *Storage
+	storage    presentationstorage.Storage
```
decentralgabe, line 33: should this still be a ptr?

pkg/service/presentation/service.go
```diff
@@ -308,3 +295,88 @@ func (s Service) ListDefinitions(ctx context.Context) (*model.ListDefinitionsRes
...
+	pd, err := s.storage.GetDefinition(ctx, request.PresentationDefinitionID)
+	if err != nil {
+		return nil, errors.Wrap(err, "getting presentation definition")
+	}
+	if pd == nil {
+		return nil, errors.Errorf("presentation definition %q is nil", request.PresentationDefinitionID)
+	}
+
+	requestID := uuid.NewString()
+	token, err := jwt.NewBuilder().Claim("presentation_definition", pd.PresentationDefinition).
```
decentralgabe, line 314: we should use https://github.com/TBD54566975/ssi-service/blob/main/internal/keyaccess/jwt.go

if we don't have exactly the right method - let's add it

pkg/service/presentation/storage.go
```diff
@@ -20,19 +20,51 @@ import (
...
 type Storage struct {
 	db storage.ServiceStorage
 }
 
+func (ps *Storage) StoreRequest(ctx context.Context, request prestorage.StoredRequest) error {
+	id := request.ID
+	if id == "" {
+		err := errors.New("could not store presentation request without an ID")
+		logrus.WithError(err).Error()
+		return err
```
decentralgabe, line 35: `sdkutil.LoggingErrorMsg`

pkg/service/presentation/storage.go
```diff
@@ -20,19 +20,51 @@ import (
...
+	}
+	jsonBytes, err := json.Marshal(request)
+	if err != nil {
+		errMsg := fmt.Sprintf("could not store presentation request: %s", id)
+		logrus.WithError(err).Error(errMsg)
+		return errors.Wrapf(err, errMsg)
```
decentralgabe, line 41: `sdkutil.LoggingErrorMsg`

pkg/service/presentation/storage/storage.go
```diff
@@ -12,17 +14,21 @@ import (
...
 }
 
 type DefinitionStorage interface {
-	StoreDefinition(schema StoredDefinition) error
-	GetDefinition(id string) (*StoredDefinition, error)
-	DeleteDefinition(id string) error
+	StoreDefinition(ctx context.Context, presentation StoredDefinition) error
+	GetDefinition(ctx context.Context, id string) (*StoredDefinition, error)
+	DeleteDefinition(ctx context.Context, id string) error
+	ListDefinitions(ctx context.Context) ([]StoredDefinition, error)
```
decentralgabe, line 31: our other APIs use `GetDefinitions` instead of `List` - I'm ok with list but let's keep them all in line

pkg/service/presentation/storage/storage.go
```diff
@@ -38,10 +44,26 @@ func (s StoredSubmission) FilterVariablesMap() map[string]any {
...
+	UpdateSubmission(ctx context.Context, id string, approved bool, reason string, submissionID string) (StoredSubmission, opstorage.StoredOperation, error)
 }
 
 var ErrSubmissionNotFound = errors.New("submission not found")
+
+type StoredRequest struct {
+	ID                        string   `json:"id"`
+	Audience                  []string `json:"audience"`
+	Expiration                string   `json:"expiration"`
+	IssuerDID                 string   `json:"issuerID"`
```
decentralgabe, line 59:
```suggestion
	IssuerDID                 string   `json:"issuerId"`
```

pkg/server/router/presentation.go
andresuribe87: Author is not part of a Presentation Definition. Adding things that aren't part of the spec makes it confusing for users.

Additionally, author moved to be the `IssuerID` field in the Presentation Request.

pkg/server/router/presentation.go
andresuribe87: I think it's less secure to have it off by default. I don't know what will be more common tbh, but as we have them, presentation requests are meant to be specific to an audience, and that means it's likely short lived.

pkg/service/presentation/model/model.go
andresuribe87: It's a more RESTful pattern. Instead of doing an additional GET, you send back what the GET will return.

pkg/service/presentation/service.go
andresuribe87: No. `presentationstorage.Storage` is an interface.

pkg/service/presentation/service.go
andresuribe87: I disagree. That package doesn't have access to the `keystore`, and I don't think it should.

The `Sign` method of the `keystore` is calling keyaccess under the hood.

pkg/service/presentation/storage.go
andresuribe87: Done

pkg/service/presentation/storage.go
andresuribe87: Done

pkg/service/presentation/storage/storage.go
andresuribe87: Done

pkg/service/presentation/storage/storage.go
```diff
@@ -38,10 +44,26 @@ func (s StoredSubmission) FilterVariablesMap() map[string]any {
...
+	KID                       string   `json:"kid"`
```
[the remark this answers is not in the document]
andresuribe87: Done

pkg/service/presentation/model/model.go
andresuribe87: Done

pkg/service/presentation/storage/storage.go
andresuribe87: Created https://github.com/TBD54566975/ssi-service/issues/449

pkg/server/router/presentation.go
decentralgabe: less secure how? `exp` is an optional claim so seems odd to make it required

pkg/service/presentation/model/model.go
decentralgabe: I'm ok with this but it's not in line with our other APIs

pkg/service/presentation/service.go
decentralgabe: it doesn't need to - here is how it's used https://github.com/TBD54566975/ssi-service/blob/12083e75bcac0d8e18b03bd4d7957fe83873a964/pkg/service/credential/service.go#L345

I'd like to stay consistent

andresuribe87: I'm not following the suggestion you're making. Would you mind clarifying what pieces to move where?

pkg/server/router/presentation.go
andresuribe87: It reduces the time span during which an attacker can intercept the request and impersonates the receiver of the submission.

decentralgabe: Unless they can steal or break a key this isn't a realistic threat. What would most likely happen is the requester would have to create and manage a new PR which is more overhead

pkg/service/presentation/service.go
decentralgabe: I would remove the jwt builder in favor of some code like

```go
gotKey, err := s.keyStore.GetKey(ctx, keystore.GetKeyRequest{ID: request.IssuerKID})
if err != nil {
	return nil, sdkutil.LoggingErrorMsgf(err, "getting key for signing request<%s>", request.IssuerKID)
}
keyAccess, err := keyaccess.NewJWKKeyAccess(request.IssuerKID, gotKey.ID, gotKey.Key)
	if err != nil {
		return nil, errors.Wrapf(err, "creating key access for signing request with key<%s>", gotKey.ID)
	}
requestToken, err := keyAccess.Sign(request)
if err != nil {
	return nil, errors.Wrapf(err, "could not sign request with key<%s>", gotKey.ID)
}
```

decentralgabe (approved):

decentralgabe (approved):

decentralgabe (approved):

decentralgabe (approved):

pkg/server/router/presentation.go
andresuribe87: Done

PR #450, opened 2023-05-19
Add integration test for webhook and fix broken middleware setup. 

andresuribe87:
The test added failed before, and passed after.


config/dev.toml
```diff
@@ -16,7 +16,7 @@ read_timeout = 5000000000
 write_timeout = 5000000000
 shutdown_timeout = 5000000000
 
-log_location = "log"
+log_location = ""
```
decentralgabe, line 19: dev is a local build and we have a log folder where this is works and useful - why remove?

andresuribe87: I found it more useful to log to stdout instead of having to tail a new log location each time.

decentralgabe: is it possible to do both with an additional setting instead of removing this?

decentralgabe: it's been valuable to view logs even after the server dies and do comparisons

decentralgabe:
looks like this is failing in the integration test `/home/runner/work/ssi-service/ssi-service/integration/didweb_integration_test.go:50`

we may want to set the logs in the `integration` pkg to `debug` instead of info it's pretty hard to see what's going on

config/dev.toml
decentralgabe: this could work https://gin-gonic.com/docs/examples/write-log/

decentralgabe (approved):

config/dev.toml
andresuribe87: Created https://github.com/TBD54566975/ssi-service/issues/462

nitro-neal (approved):
Great job! I had something similar on my box but I couldn't get it working all the way. SSI service tried to hit my test server and it would never get through.

Awesome job figuring it out!

PR #465, opened 2023-05-22
update ion url

decentralgabe:
change to our new host


nitro-neal (approved):
There may be a unit test in here that references the old one, maybe do a code wide search for the old one

decentralgabe:
@nitro-neal got all of them there is a test in the sdk which I'll cover separately

PR #466, opened 2023-05-22
rename issuing to issuance

decentralgabe:
keep this in line with the rest of our APIs


nitro-neal (approved):
https://github.com/TBD54566975/ssi-playground/blob/main/ssi-service-compose/config.toml

fyi breaking changes to the toml will need to be updated to playground or it may not spin up (It uses the latest ssi image)

We probably need to find a better way to deal with that, probably just version a docker image

decentralgabe:
https://github.com/TBD54566975/ssi-playground/pull/10

PR #468, opened 2023-05-23
Publish webhooks in parallel

### LINKED ISSUE #463: Calls to registered webhook endpoint should be done in parallel
The following line should be done in parallel. There shouldn't be a reason why one bad webhook should affect the rest.

 https://github.com/TBD54566975/ssi-service/blob/7047be55d3d57ca5e4ee4919126e4acdeb275444/pkg/service/webhook/service.go#L208

andresuribe87:
Fixed #463


pkg/service/webhook/service.go
```diff
@@ -205,10 +207,15 @@ func (s Service) PublishWebhook(noun Noun, verb Verb, payloadReader io.Reader) {
 			continue
 		}
 
-		if err = s.post(ctx, url, string(postJSONData)); err != nil {
-			logrus.WithError(err).Errorf("posting payload to %s", url)
-		}
+		wg.Add(1)
+		go func(url, data string) {
+			defer wg.Done()
+			if err := s.post(ctx, url, data); err != nil {
```
nitro-neal, line 213: Very cool.

If it ever becomes an issue and we have 1000 urls to post to eventually we can do batching / retry logic. For now this is great though

nitro-neal (approved):

decentralgabe (approved):

PR #472, opened 2023-05-23
Add public key to GetKeyDetails

### LINKED ISSUE #471: Return public keys in the keystore API
We're generating and storing private keys, we should expose the public keys.

The public keys can be stored unencrypted. No need to reed the decrypt the private key to get the value.

andresuribe87:
# Overview
Fixes #471 

# Description
This adds a JWK field to the response. 
The public key is stored in the DB without being encrypted. It's stored in a separate namespace. The downside is that there are two DB calls when fetching the details. This could be reduced, but I think the separation of concerns is worth it. 

# How Has This Been Tested?
Added a test to check that the returned value is correct.


pkg/server/router/keystore.go
```diff
@@ -103,7 +104,13 @@ type GetKeyDetailsResponse struct {
...
 	Type       crypto.KeyType `json:"type,omitempty"`
 	Controller string         `json:"controller,omitempty"`
-	CreatedAt  string         `json:"createdAt,omitempty"`
+
+	// Represents the time at which the key was created. Encoded according to RFC3339.
+	CreatedAt string `json:"createdAt,omitempty"`
+
+	// The public key in JWK format according to RFC7517. This public key is associated with the private
+	// key with the associated ID.
+	PublicKeyJWK jwx.PublicKeyJWK `json:"publicKeyJwk"`
```
decentralgabe, line 113: we should be careful here since not all key types support jwk. for example we don't have a way (yet) to represent BLS curves as JWKs 

this is probably fine for now since we don't have BBS+ in the service but this could be JWK or PublicKeyBase58

decentralgabe (approved):

PR #474, opened 2023-05-24
expose storage object in instantiateServices for later use

andorsk:
# Overview

This allows the storage object to be reachable from outside the instantiateServices's function. There are good reasons to do this. It allows the storage provider to be reused on services that might extend the ssi service,  which is a reasonable use case IMO (we are doing it at Benri). Bolt, in particular, is finicky with how it interacts with storage objects. If you concurrently open two bolt providers for the same file on the same service, it seems the first open locks the second ( at least in my experience ), so reusing the storage provider for other functions requires this to be exposed.

### DIFF
--- a/pkg/service/service.go
+++ b/pkg/service/service.go
@@ -30,6 +30,7 @@ type SSIService struct {
 	Presentation *presentation.Service
 	Operation    *operation.Service
 	Webhook      *webhook.Service
+	storage      storage.ServiceStorage
 }
 
 // InstantiateSSIService creates a new instance of the SSIS which instantiates all services and their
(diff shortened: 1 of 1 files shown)


decentralgabe (approved):

decentralgabe:
@andorsk I'm OK with this PR, but I'm not sure why it's beneficial to expose storage after instantiation as opposed to passing it somewhere else upon creation

andorsk:
@decentralgabe b/c if you build the server using NewSSIServer, the storage is hidden under some private methods: https://github.com/TBD54566975/ssi-service/blob/main/pkg/service/service.go#L81, so it's unreachable. You would have to override the instantiateServices, and by association https://github.com/TBD54566975/ssi-service/blob/main/pkg/service/service.go#L37,  to get a similar result.

PR #475, opened 2023-05-24
Fix /v1/presentations paths

decentralgabe:


nitro-neal (approved):

decentralgabe:
covering here instead https://github.com/TBD54566975/ssi-service/pull/477

PR #481, opened 2023-05-24
Implemented pagination in the storage layer

andresuribe87:
# Overview
Introduces the `ReadPage` method to the `ServiceStorage` interface. 

# Description
Implements the new method in both `bolt` and `redis`. This is the first step in supporting pagination of elements. A future PR can use this new interface to support paginated endpoints. 

# How Has This Been Tested?
Added a unit test. Also added a test for `WriteMany`, which turns out was broken.


decentralgabe (approved):
nice!

PR #489, opened 2023-05-25
List Credentials without params

nitro-neal:
This change allows you to list all credentials without a query param. The query params are now optional.

image


pkg/server/router/credential.go
```diff
@@ -438,9 +442,23 @@ func (cr CredentialRouter) ListCredentials(c *gin.Context) {
...
+	gotCredentials, err := cr.service.ListCredentials(c)
+	if err != nil {
+		errMsg := fmt.Sprintf("could not get credentials")
+		framework.LoggingRespondErrWithMsg(c, err, errMsg, http.StatusInternalServerError)
+		return
+	}
+
+	resp := ListCredentialsResponse{Credentials: gotCredentials.Credentials}
+	framework.Respond(c, resp, http.StatusOK)
+	return
```
andresuribe87, line 459:
```suggestion
```

pkg/server/server_credential_test.go
```diff
@@ -356,6 +356,63 @@ func TestCredentialAPI(t *testing.T) {
 		assert.Equal(tt, resp.Credential.CredentialSchema.ID, getCredsResp.Credentials[0].Credential.CredentialSchema.ID)
 	})
 
+	t.Run("Test Get Credential No Param", func(tt *testing.T) {
```
andresuribe87, line 359:
```suggestion
	t.Run("Get Credential No Param", func(tt *testing.T) {
```

pkg/service/credential/service.go
```diff
@@ -431,6 +431,30 @@ func (s Service) GetCredential(ctx context.Context, request GetCredentialRequest
 	return &response, nil
 }
 
+func (s Service) ListCredentials(ctx context.Context) (*ListCredentialsResponse, error) {
+	logrus.Debugf("listing credential(s) ")
+
+	gotCreds, err := s.storage.GetCredentials(ctx)
```
andresuribe87, line 437: Should we stick to `ListCredentials`?

pkg/service/credential/storage.go
```diff
@@ -346,6 +346,36 @@ func (cs *Storage) getCredential(ctx context.Context, id string, namespace strin
...
+	keys, err := cs.db.ReadAllKeys(ctx, credentialNamespace)
+	if err != nil {
+		return nil, sdkutil.LoggingErrorMsgf(err, "could not read credential storage")
+	}
+
+	var storedCreds []StoredCredential
+	for _, key := range keys {
+		credBytes, err := cs.db.Read(ctx, credentialNamespace, key)
+		if err != nil {
+			logrus.WithError(err).Errorf("could not read credential with key: %s", key)
```
andresuribe87, line 362: Is the intention to skip this? I think we should return an error.

pkg/service/credential/storage.go
```diff
@@ -346,6 +346,36 @@ func (cs *Storage) getCredential(ctx context.Context, id string, namespace strin
...
+		} else {
+			var cred StoredCredential
+			if err = json.Unmarshal(credBytes, &cred); err != nil {
+				logrus.WithError(err).Errorf("unmarshalling credential with key: %s", key)
```
andresuribe87, line 366: Same here.

pkg/service/credential/storage.go
```diff
@@ -346,6 +346,36 @@ func (cs *Storage) getCredential(ctx context.Context, id string, namespace strin
...
+			}
+			storedCreds = append(storedCreds, cred)
+		}
+	}
+
+	if len(storedCreds) == 0 {
+		logrus.Warnf("no credentials able to be retrieved")
```
andresuribe87, line 373: Why warn? This may be a valid use case :)

pkg/service/credential/storage.go
decentralgabe: yeah as the comment for the method says this is greedy - we warn we couldn't get one but return what we have. in case of data corruption we wouldn't want to eliminate the good data

we may want to amend this to return both data and an error

pkg/service/credential/storage.go
decentralgabe: could change to info

pkg/service/credential/storage.go
andresuribe87: Either way is fine. We just need to document this so it's reflected in our API.

andresuribe87 (approved):

PR #504, opened 2023-05-31
Added List presentation requests method

### LINKED ISSUE #479: PresentationAPI: Add list requests endpoint
We support create, get, and delete, but not an endpoint to list all requests

andresuribe87:
# Overview
Fixes #479 

# Description
Added a new endpoint with docs and tests. 

# How Has This Been Tested?
Added unit tests.


pkg/service/presentation/service.go
```diff
@@ -27,6 +27,8 @@ import (
 	"github.com/tbd54566975/ssi-service/pkg/storage"
 )
 
+const presentationRequestNamespace = "presentation_request"
```
decentralgabe, line 30: have we gone with snake case for other namespaces?

andresuribe87: It's been a bit all over the place, created https://github.com/TBD54566975/ssi-service/issues/506 to improve.

PR #505, opened 2023-06-01
Support credentialschema2023

### LINKED ISSUE #501: Support CredentialSchema2023
https://w3c.github.io/vc-json-schema/#credentialschema2023
- Support packaging schemas as VCs using Credential Schema 2023
- Need to determine whether to introduce a cyclic dependency between the schema and credentials service; or, have schema create its own credentials

thereisnogabe:
Fixes https://github.com/TBD54566975/ssi-service/issues/501


pkg/server/router/schema.go
```diff
@@ -39,20 +39,23 @@ type CreateSchemaRequest struct {
...
-	// Issuer represents the DID of the issuer for the schema if it's signed. Required if sign is true.
-	// Issuer string `json:"issuer,omitempty"`
-	// IssuerKID represents the KID of the issuer's private key to sign the schema. Required if sign is true.
-	// IssuerKID string `json:"issuerKid,omitempty"`
+	// Issuer represents the DID of the issuer for the schema if it's signed. Required if intending to sign the
+	// schema as a credential using CredentialSchema2023.
+	Issuer string `json:"issuer,omitempty"`
+	// IssuerKID represents the KID of the issuer's private key to sign the schema. Required if intending to sign the
+	// schema as a credential using CredentialSchema2023.
+	IssuerKID string `json:"issuerKid,omitempty"`
```
andresuribe87, line 47: I suggest moving these two fields into a separate struct. The presence of the struct would then signal the intention of the user. When present, then both fields should be populated, and the intention of the user is clear that they want a CredentialSchema2023. When absent, then the user wants to do a JsonSchema2023.

pkg/server/router/schema.go
```diff
@@ -39,20 +39,23 @@ type CreateSchemaRequest struct {
...
 }
 
 type CreateSchemaResponse struct {
-	ID               string               `json:"id"`
-	Schema           schemalib.JSONSchema `json:"schema"`
-	CredentialSchema *keyaccess.JWT       `json:"credentialSchema,omitempty"`
+	ID   string                     `json:"id"`
+	Type schemalib.VCJSONSchemaType `json:"type" validate:"required"`
```
andresuribe87, line 52: Comments for these two please. Let's keep the habit of giving love to our docs :)

pkg/service/schema/service.go
```diff
@@ -116,14 +117,84 @@ func (s Service) CreateSchema(ctx context.Context, request CreateSchemaRequest)
...
+	return &CreateSchemaResponse{
+		ID:               schemaID,
+		Type:             storedSchema.Type,
+		Schema:           storedSchema.Schema,
+		CredentialSchema: storedSchema.CredentialSchema,
+	}, nil
+}
+
+// createCredentialSchema creates a credential schema, and signs it with the issuer's key and kid
+func (s Service) createCredentialSchema(ctx context.Context, jsonSchema schema.JSONSchema, id, issuer, issuerKID string) (*keyaccess.JWT, error) {
```
andresuribe87, line 154:
```suggestion
func (s Service) createCredentialSchema(ctx context.Context, jsonSchema schema.JSONSchema, schemaURI, issuer, issuerKID string) (*keyaccess.JWT, error) {
```

pkg/service/schema/service.go
```diff
@@ -169,10 +245,29 @@ func (s Service) DeleteSchema(ctx context.Context, request DeleteSchemaRequest)
...
+		if err != nil {
+			return nil, "", sdkutil.LoggingErrorMsg(err, "converting credential schema from jwt to credential map")
+		}
+		credSubjectBytes, err := json.Marshal(cred.CredentialSubject)
+		if err != nil {
+			return nil, "", errors.Wrap(err, "error marshalling credential subject")
+		}
+		var s schema.JSONSchema
+		if err = json.Unmarshal(credSubjectBytes, &s); err != nil {
+			return nil, "", errors.Wrap(err, "error unmarshalling credential subject")
```
andresuribe87, line 267:
```suggestion
			return nil, "", errors.Wrap(err, "unmarshalling credential subject")
```

andresuribe87 (approved):

pkg/service/schema/service.go
```diff
@@ -116,14 +117,84 @@ func (s Service) CreateSchema(ctx context.Context, request CreateSchemaRequest)
 		jsonSchema[schema.JSONSchemaDescriptionProperty] = request.Description
 	}
 
-	// TODO(gabe) support signing credential schemas
-	// create schema
-	storedSchema := StoredSchema{ID: schemaID, Schema: jsonSchema}
+	// if the schema is a credential schema, the credential's id is a fully qualified URI
+	// if the schema is a JSON schema, the schema's id is a fully qualified URI
+	schemaID := uuid.NewString()
+	schemaURI := strings.Join([]string{s.Config().ServiceEndpoint, schemaID}, "/")
```
andresuribe87, line 123: Actually I think there is something off here. When request is for a `CredentialSchemar2023`, there should be a URL communicated that can be dereferenced to get the credential. We don't have that right now.

andresuribe87 (changes requested):

pkg/server/router/schema.go
decentralgabe: good idea

pkg/server/router/schema.go
```diff
@@ -39,20 +39,23 @@ type CreateSchemaRequest struct {
...
+
+	// Schema is the JSON schema for the credential, returned when the type is JsonSchema2023
+	Schema *schemalib.JSONSchema `json:"schema,omitempty"`
+
+	// CredentialSchema is the JWT schema for the credential, returned when the type is CredentialSchema2023
+	CredentialSchema *keyaccess.JWT `json:"credentialSchema,omitempty"`
```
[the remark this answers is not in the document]
decentralgabe: done

pkg/service/schema/service.go
decentralgabe: this should go through the `v1/schemas/id` endpoint though?

pkg/server/server_schema_test.go
```diff
@@ -53,6 +57,73 @@ func TestSchemaAPI(t *testing.T) {
 		// since the id is generated, we need to manually override it
 		schemaRequest.Schema[schema.JSONSchemaIDProperty] = resp.Schema.ID()
 		assert.JSONEq(tt, schemaRequest.Schema.String(), resp.Schema.String())
+		assert.Equal(tt, schema.JSONSchema2023Type, resp.Type)
+		assert.Empty(tt, resp.CredentialSchema)
+		assert.NotEmpty(tt, resp.Schema)
+	})
+
+	t.Run("Test Create CredentialSchema2023 Schema", func(tt *testing.T) {
```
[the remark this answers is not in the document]
decentralgabe: probably but I'd rather do a big refactor later

pkg/service/schema/service.go
```diff
@@ -116,14 +117,84 @@ func (s Service) CreateSchema(ctx context.Context, request CreateSchemaRequest)
...
+		return nil, sdkutil.LoggingErrorMsgf(err, "could not set subject: %+v", subject)
+	}
+	if err := builder.SetIssuanceDate(time.Now().Format(time.RFC3339)); err != nil {
+		return nil, sdkutil.LoggingErrorMsg(err, "could not set credential schema issuance date")
+	}
+	cred, err := builder.Build()
+	if err != nil {
+		return nil, sdkutil.LoggingErrorMsg(err, "could not build credential schema")
+	}
+	return s.signCredentialSchema(ctx, *cred, issuer, issuerKID)
```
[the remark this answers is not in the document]
decentralgabe: I had to revert this because it wasn't constructing the VC correctly. need the call to 

`keyAccess.SignVerifiableCredetial`

pkg/service/schema/service.go
andresuribe87: Create #508 to track

andresuribe87 (approved):

PR #509, opened 2023-06-01
Credential manifests can be created with a definition ID

### LINKED ISSUE #491: Allow for specifying the presentation definition only by id during manifest creation
When creating a credential manifest, we should allowing users to specify the presentation definition by id.

andresuribe87:
# Overview
Fixes #491 

# Description
This allows users to specify the ID of an already created presentation definition when created a credential manifest. 

# How Has This Been Tested?
Added tests for the new functionality.
Existing test still pass.


pkg/server/router/manifest.go
```diff
@@ -44,14 +44,38 @@ func NewManifestRouter(s svcframework.Service) (*ManifestRouter, error) {
...
+	// DID that identifies who the issuer of the credential(s) will be.
+	// Required.
+	IssuerDID string `json:"issuerDid" validate:"required"`
+
+	// The KID of the private key that will be used when signing issued credentials.
+	// Required.
+	IssuerKID string `json:"issuerKid" validate:"required"`
+
+	// Human-readable name the Issuer wishes to be recognized by.
+	// Optional.
```
nitro-neal, line 64: Having the optional comment is nice, but a little jank. I wonder if there is a way to specify it in the `json:` part

I guess omitempty is this, 'if its missing will not add it to the json obect' so anything that has omitempty is optional.

Its find though, I think the comment is good

pkg/service/manifest/model/model.go
```diff
@@ -179,3 +178,11 @@ type Request struct {
 	// This is an output only field.
 	CredentialManifestJWT keyaccess.JWT `json:"credentialManifestJwt"`
 }
+
+type PresentationDefinitionRef struct {
+	// id of the presentation definition created with PresentationDefinitionAPI. Must be empty if `value` is present.
+	ID *string `json:"presentationDefinitionId"`
+
+	// value of the presentation definition to use. Must be empty if `id` is present.
```
nitro-neal, line 186: Is this enforced? if I put a id and presentationDefinition it will throw an error right?

pkg/service/manifest/service.go
```diff
@@ -157,7 +160,24 @@ func (s Service) CreateManifest(ctx context.Context, request model.CreateManifes
 		)
 	}
 	if request.PresentationDefinition != nil {
-		if err := builder.SetPresentationDefinition(*request.PresentationDefinition); err != nil {
+		var pd *exchange.PresentationDefinition
+		if request.PresentationDefinition.ID != nil && request.PresentationDefinition.PresentationDefinition != nil {
+			return nil, errors.New(`only one of "id" and "value" can be provided`)
```
nitro-neal, line 165: nice its enforced

pkg/service/manifest/service.go
```diff
@@ -157,7 +160,24 @@ func (s Service) CreateManifest(ctx context.Context, request model.CreateManifes
 		)
 	}
 	if request.PresentationDefinition != nil {
```
nitro-neal, line 162: should this be called PresentationDefinitionRef?

nitro-neal (approved):
Nice job

pkg/server/router/manifest.go
andresuribe87: We need to be able to specify this optionality at the API level. It would be awesome if we could simply use a type and it get automatically populated, but our tooling doesn't support it :(

pkg/service/manifest/service.go
andresuribe87: yup!

pkg/service/manifest/service.go
andresuribe87: Good call, done.

PR #518, opened 2023-06-07
Support pagination for DIDs

andresuribe87:
# Overview
This is the first of many PRs that will address #349. It focuses on supporting pagination for DIDs.

# Description
The API defined accepts `pageSize` and `pageToken` as query parameters. The List response object will contain a `nextPageToken` field which can be passed to subsequent requests. It's also backwards compatible. When the query parameters are not provided, then the previous behavior is maintained. 

# How Has This Been Tested?
Unit tests.


pkg/server/router/did.go
```diff
@@ -263,21 +279,83 @@ func (dr DIDRouter) ListDIDsByMethod(c *gin.Context) {
...
+		query := pageTokenQuery(c)
+		if !reflect.DeepEqual(pageTokenValues, query) {
+			logrus.Warnf("expected query from token to be equal to query from request. token: %v\nrequest%v", pageTokenValues, query)
+			framework.LoggingRespondErrMsg(c, "page token must be for the same query", http.StatusBadRequest)
+			return
+		}
+		getDIDsRequest.PageToken = &pageToken.NextPageToken
+	}
+
+	listResp, err := dr.service.ListDIDsByMethod(c, getDIDsRequest)
```
nitro-neal, line 326: This can probably be put in a util shared by all routes so they can share 'page processing' code

pkg/service/did/ion.go
```diff
@@ -47,6 +47,8 @@ type ionHandler struct {
 	keyStore *keystore.Service
 }
 
+var _ MethodHandler = (*ionHandler)(nil)
```
nitro-neal, line 50: Maybe put a comment that CHAD told me about:
This line doesn't actually do anything at runtime; it's purely for compile-time checking. If ionHandler doesn't implement all methods required by MethodHandler, you'll get a compile-time error.

pkg/service/did/service.go
```diff
@@ -202,7 +202,18 @@ func (s *Service) ListDIDsByMethod(ctx context.Context, request ListDIDsRequest)
 	if request.Deleted {
 		return handler.ListDeletedDIDs(ctx)
```
nitro-neal, line 203: List Deleted Dids wont be paginated? (that is probably ok)

nitro-neal (approved):
Looks good! 
Maybe adding a quick integration test would be good too.

I hope we can abstract a lot of this in a shared package so there is not that much duplicate code.

Could -almost- do a middleware where it would abstract a lot of this away, but probably hard / confusing to do

pkg/server/router/did.go
```diff
@@ -263,21 +279,83 @@ func (dr DIDRouter) ListDIDsByMethod(c *gin.Context) {
...
 	}
 
-	resp := ListDIDsByMethodResponse{DIDs: gotDIDs.DIDs}
+	resp := ListDIDsByMethodResponse{
+		DIDs: listResp.DIDs,
+	}
+	if listResp.NextPageToken != "" {
+		tokenQuery := pageTokenQuery(c)
+		pageToken := PageToken{
+			EncodedQuery:  tokenQuery.Encode(),
```
nitro-neal, line 339: Can people 'hack this' and encode other random query data to get like 'the wrong page' lol

pkg/server/router/did.go
andresuribe87: Absolutely! I'm going to do that in a follow up PR where I enable it for another List endpoint.

andresuribe87:
> Looks good! Maybe adding a quick integration test would be good too.

Good point. I'll get to it. 

> I hope we can abstract a lot of this in a shared package so there is not that much duplicate code.
> 
> Could -almost- do a middleware where it would abstract a lot of this away, but probably hard / confusing to do

Absolutely! Will come in next PRs. Want to make sure to get something that works before generalizing.

doc/swagger.yaml
```diff
@@ -1514,6 +1514,10 @@ definitions:
         items:
           $ref: '#/definitions/did.Document'
         type: array
+      nextPageToken:
```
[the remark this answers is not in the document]
andresuribe87: It could, but I find `nextPageToken` to be more informative to consumers of the API.

Issue #347, opened 2023-03-24
Support for Batch Operations


decentralgabe:
Support for batches, e.g.:
- Batch issuance of credentials
- Batch status changes (revocation)
- Batch credential refresh
- ... more!

PR #520, opened 2023-06-07
Batch Create Implementation

andresuribe87:
# Overview
This is part of #347. It implements batch create credentials.

# Description
Note that the creation is atomic. No partial creations are allowed. There is also a limit to how many requests can be included in a batch. 

# How Has This Been Tested?
Unit tests.
Added an integration test.


pkg/server/router/credential.go
```diff
@@ -37,6 +37,65 @@ func NewCredentialRouter(s svcframework.Service) (*CredentialRouter, error) {
...
+//	@Router			/v1/credentials/batchCreate [put]
+func (cr CredentialRouter) BatchCreateCredentials(c *gin.Context) {
+	invalidCreateCredentialRequest := "invalid batch create credential request"
+	var batchRequest BatchCreateCredentialsRequest
+	if err := framework.Decode(c.Request, &batchRequest); err != nil {
+		framework.LoggingRespondErrWithMsg(c, err, invalidCreateCredentialRequest, http.StatusBadRequest)
+		return
+	}
+
+	const batchCreateMaxItems = 1000
```
nitro-neal, line 78: could be a config item

pkg/server/server.go
```diff
@@ -185,6 +185,7 @@ func CredentialAPI(rg *gin.RouterGroup, service svcframework.Service, webhookSer
 	// Credentials
 	credentialAPI := rg.Group(CredentialsPrefix)
 	credentialAPI.PUT("", middleware.Webhook(webhookService, webhook.Credential, webhook.Create), credRouter.CreateCredential)
+	credentialAPI.PUT("/batchCreate", middleware.Webhook(webhookService, webhook.Credential, webhook.BatchCreate), credRouter.BatchCreateCredentials)
```
nitro-neal, line 188: This will fire once and not 1000 times right?
the output may be too large for a http POST req... 

also I wonder if someone is just listenting to /CreateCredential if it will fire 1000 times

integration/credential_manifest_integration_test.go
```diff
@@ -102,6 +102,44 @@ func TestCreateVerifiableCredentialIntegration(t *testing.T) {
 	SetValue(credentialManifestContext, "credentialJWT", credentialJWT)
 }
 
+func TestBatchCreateCredentialsIntegration(t *testing.T) {
```
nitro-neal, line 105: Lets make one with 1000!

pkg/server/server.go
nitro-neal: Seems like only if they subscribe to BatchCreate it will fire and not Create-Credential

But yea the post with the data will be too large.

Speaking of that, if we really make an integ test for 1000 creds the golang part may fail, like we may have a max of 10mb per response default config somewhere or something, worth testing!

pkg/server/router/credential.go
andresuribe87: Done

pkg/server/router/credential.go
```diff
@@ -37,6 +37,65 @@ func NewCredentialRouter(s svcframework.Service) (*CredentialRouter, error) {
...
+//	@Summary		Batch Create Credentials
+//	@Description	Create a batch of verifiable credentials.
+//	@Tags			CredentialAPI
+//	@Accept			json
+//	@Produce		json
+//	@Param			request	body		BatchCreateCredentialsRequest	true	"The batch requests"
+//	@Success		201		{object}	BatchCreateCredentialsResponse
+//	@Failure		400		{string}	string	"Bad request"
+//	@Failure		500		{string}	string	"Internal server error"
+//	@Router			/v1/credentials/batchCreate [put]
```
decentralgabe, line 69: gut level this feels off and not very resutful

what about `v1/credentials/batch` ?

pkg/server/router/credential.go
```diff
@@ -37,6 +37,65 @@ func NewCredentialRouter(s svcframework.Service) (*CredentialRouter, error) {
...
+func (cr CredentialRouter) BatchCreateCredentials(c *gin.Context) {
+	invalidCreateCredentialRequest := "invalid batch create credential request"
+	var batchRequest BatchCreateCredentialsRequest
+	if err := framework.Decode(c.Request, &batchRequest); err != nil {
+		framework.LoggingRespondErrWithMsg(c, err, invalidCreateCredentialRequest, http.StatusBadRequest)
+		return
+	}
+
+	batchCreateMaxItems := cr.service.Config().BatchCreateMaxItems
```
decentralgabe, line 78: for the future we should have a method such as `GetOrDefault` on each property

pkg/server/server_credential_test.go
```diff
@@ -26,6 +26,111 @@ import (
...
+					Subject:   "did:abc:789",
+					Data: map[string]any{
+						"firstName": "Lemony",
+						"lastName":  "Snickets",
+					},
+					Expiry:    time.Now().Add(12 * time.Hour).Format(time.RFC3339),
+					Revocable: true,
+				},
+			},
+		}
```
decentralgabe, line 69: nl

pkg/server/router/credential.go
andresuribe87: If you're referring to the config struct, it already has a default value annotation. What would the `GetOrDefault` method do, different from returning the `BatchCreateMaxItems` value?

pkg/server/router/credential.go
decentralgabe: still should be clear since it's namespaced to credentials and follows PUT for rest semantics

decentralgabe: yes, on quick google:
- https://www.mscharhag.com/api-design/bulk-and-batch-operations
- https://www.codementor.io/blog/batch-endpoints-6olbjay1hd
note this one shows a google API `POST https://www.googleapis.com/batch/drive/v3`
- https://www.ibm.com/docs/en/was-liberty/base?topic=liberty-batch-rest-api
- https://www.servicenow.com/community/now-platform-articles/an-introduction-to-the-batch-rest-api-endpoint-api-now-batch/ta-p/2317136

andresuribe87: the examples all have different approaches. None use a `<resource>/batch` approach. Are you suggesting to have a top level `v1/batch` endpoint? 
* https://www.mscharhag.com/api-design/bulk-and-batch-operations suggest using a `v1/batch` endpoint, without the resources it's acting on. It supports batching of many things. 
* https://www.codementor.io/blog/batch-endpoints-6olbjay1hd also showcases an example to an endpoint that has the verb on it `/api/v2/users/create_or_update_many.json`. The drive endpoint uses the approach of having a 
* https://www.ibm.com/docs/en/was-liberty/base?topic=liberty-batch-rest-api uses `batch` to namespace all batch requests `/ibm/api/batch/jobinstances`
* https://www.servicenow.com/community/now-platform-articles/an-introduction-to-the-batch-rest-api-endpoint-api-now-batch/ta-p/2317136 again usese the `batch` namespace approach.

decentralgabe: I'm ok with the batch namespace approach, my main gripe is that `batch` is more of a resource than `batchCreate` which is an RPC, and we should be as RESTful as possible

decentralgabe: another option is no batch endpoint, but a batch flag or request parameter

PR #524, opened 2023-06-08
Remove unneeded swaggo generated file. 

### LINKED ISSUE #522: Merging a PR that contains doc changes causes merge conflicts
The `doc/docs.go` file that is generated by swaggo hass all the swagger doc in a single line. When any doc is updates, this single line changes since git isn't smart enough to recognize how to edit it. This is annoying for development as merge conflicts need to be resolved manually. We can do better. Some alternatives
1. Do not generate the go file and read from the generated swagger file. 
2. Somehow split the go file swagger string into separate lines.

andresuribe87:
# Overview
Fixes #522. This file isn't needed, so it's fine to delete. 

# Description
Fixed additional documentation issues.


pkg/server/server.go
```diff
@@ -70,10 +69,6 @@ func NewSSIServer(shutdown chan os.Signal, cfg config.SSIServiceConfig) (*SSISer
 	engine.GET(ReadinessPrefix, router.Readiness(ssi.GetServices()))
 
 	// swagger
-	doc.SwaggerInfo.Version = cfg.SVN
-	doc.SwaggerInfo.Description = cfg.Desc
-	doc.SwaggerInfo.Host = cfg.Server.APIHost
```
decentralgabe, line 75: it's useful to set these dynamically based on config

decentralgabe (changes requested):

pkg/server/router/credential.go
```diff
@@ -39,33 +39,33 @@ func NewCredentialRouter(s svcframework.Service) (*CredentialRouter, error) {
 
 type CreateCredentialRequest struct {
 	// The issuer id.
-	Issuer string `json:"issuer" validate:"required" example:"did:key:z6MkiTBz1ymuepAQ4HEHYSF1H8quG5GLVVQR3djdX3mDooWp"`
+	Issuer string `json:"issuer" validate:"required" example:"did:key:z6MkkZDjunoN4gyPMx5TSy7Mfzw22D2RZQZUcx46bii53Ex3"`
 
 	// The KID used to sign the credential.
-	IssuerKID string `json:"issuerKid" validate:"required" example:"#z6MkiTBz1ymuepAQ4HEHYSF1H8quG5GLVVQR3djdX3mDooWp"`
+	IssuerKID string `json:"issuerKid" validate:"required" example:"did:key:z6MkkZDjunoN4gyPMx5TSy7Mfzw22D2RZQZUcx46bii53Ex3#z6MkkZDjunoN4gyPMx5TSy7Mfzw22D2RZQZUcx46bii53Ex3"`
```
nitro-neal, line 45: does it work like this even on integ tests? I though we could alwasy do #z6MkkZDjunoN4gyPMx5TSy7Mfzw22D2RZQZUcx46bii53Ex3 and it would work

nitro-neal (approved):

pkg/server/server.go
andresuribe87: Can you elaborate on why this is useful? 

There are a couple of reasons why I think it's more useful to not have this here. 
1. This fixes #522
2. When setting this dynamically there ends up being a mismatch between the content of the `swagger.yaml` file and what's hosted.
3. The semantic versioning is something that's defined per PR, and shouldn't be changed in config files. In fact, it's required by the openapi spec.
4. Having this be dynamic requires a 2-step generation.

Additionally, I wasn't able to tell how these lines are achieving any visible changes.

decentralgabe: If we can figure out a way to maintain these properties - and have them up to date with our config - then no problems from me.

I believe there's another issue that is tangential to this as well. We need to distinguish between "service information" - like routes and versions - and user specified config.

doc/swagger.yaml
```diff
@@ -1035,9 +1039,8 @@ definitions:
       keyType:
         allOf:
         - $ref: '#/definitions/crypto.KeyType'
-        description: |-
-          Identifies the cryptographic algorithm family to use when generating this key.
-          One of the following: "Ed25519", "X25519", "secp256k1", "P-224","P-256","P-384", "P-521", "RSA"
```
decentralgabe, line 1040: I believe this is useful info. If not here where do we specify these types?

pkg/server/router/credential.go
andresuribe87: Yes, this is the correct format. Integration tests are picking it up directly from the DID document.

decentralgabe (approved):
nice one

PR #529, opened 2023-06-09
Create release-notifier.yml

michaelneale:
This will notify discord, and open an issue to ensure that the release notes are examined and any docs/articles are updated by devrel.


michaelneale:
cc @kirahsapong

.github/workflows/release-notifier.yml
```diff
@@ -0,0 +1,33 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Create Issue
+      uses: actions/github-script@v6
+      with:
+        github-token: ${{ secrets.DOCS_PAT }}  # Assuming 'PAT' is your Personal Access Token stored as a secret
+        script: |
+          const issueTitle = `A new release ${context.payload.release.tag_name} just cut for ${context.payload.repository.name}.`;
+          const issueBody = `Please check the api docs for compatibility with the release. Here is the [link to release notes](${context.payload.release.html_url}).`;
```
andresuribe87, line 18: api docs are updated with every commit. Would it be worth specifying which docs to look at (i think this is blog posts + educational content)?

andresuribe87 (approved):

andresuribe87:
@michaelneale Curious how you tested this?

michaelneale:
@andresuribe87 I tested it on my own repo here: https://github.com/michaelneale/sample-content/actions and annoyed people by creating fake releases and thus fake issues etc ;)

.github/workflows/release-notifier.yml
```diff
@@ -0,0 +1,33 @@
...
+
+jobs:
+  create-issue:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Create Issue
+      uses: actions/github-script@v6
+      with:
+        github-token: ${{ secrets.DOCS_PAT }}  # Assuming 'PAT' is your Personal Access Token stored as a secret
```
decentralgabe, line 15: same questions as the other repo - whose is this and how is it managed? what shall we do in case it expires and breaks the build?

michaelneale: it's mine - and if it breaks the build, we remove the notifier. Doesn't have an expiry.

decentralgabe (approved):

decentralgabe:
can this be merged @andresuribe87 @michaelneale

michaelneale:
yes if @andresuribe87 is ok with it.

Issue #523, opened 2023-06-08
[Bug] Calling `GET` Schemas returns empty `type` in response 


kirahsapong:
**Describe the bug**
After creating a schema, calling GET Schemas to get all schemas returns an array of schemas in which each of the schemas has an empty `type` property.

**To Reproduce**
1. Create a Schema. Observe the `type` property in the response. 
2. Get the schema by `id` by calling `GET /v1/schemas/{id}`. This returns the `type` value in the response as expected.
3. Get all schemas by calling `GET /v1/schemas`. This returns an empty `type` value, like `"type": ""`.

**Expected behavior**
Should return the correct `type` value

**Supporting Material**
<img width="922" alt="Screenshot 2023-06-08 at 9 21 31 AM" src="https://github.com/TBD54566975/ssi-service/assets/102400653/0962b698-239c-4a92-ab3f-e2b5597f5b2a">

PR #530, opened 2023-06-09
fixing schema type error

nitro-neal:
Fixes: https://github.com/TBD54566975/ssi-service/issues/523

ListSchemas now matches the output on GetSchema for each schema:

image

image

### DIFF
--- a/pkg/server/router/schema_test.go
+++ b/pkg/server/router/schema_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


pkg/server/router/schema_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"testing"
 
+	schema2 "github.com/TBD54566975/ssi-sdk/credential/schema"
```
andresuribe87, line 7: s/schema2/credschema/

andresuribe87 (approved):

PR #532, opened 2023-06-09
I like my deps small

andresuribe87:
No need to use something that's supported by the standard lib. Complete removal will be done when https://github.com/TBD54566975/ssi-sdk/pull/412 is merged and we upgrade.


andresuribe87:
The PR this depends on was closed as not happening, so closing this as well.

PR #537, opened 2023-06-12
Support pagination for submissions

andresuribe87:
# Overview
This is the first of many PRs that will address https://github.com/TBD54566975/ssi-service/issues/349. It focuses on supporting pagination for submissions, by building off of #518 .

# Description
This follows the same pattern as the existing one. Shared functionality was refactored into separate packages. 

# How Has This Been Tested?
Unit tests were implemented.

### DIFF
--- a/pkg/server/pagination/pagination.go
+++ b/pkg/server/pagination/pagination.go
@@ -1,0 +1,117 @@
+package pagination
+
+import (
+	"encoding/base64"
+	"fmt"
+	"net/http"
+	"net/url"
+	"reflect"
+	"strconv"
+
+	"github.com/gin-gonic/gin"
+	"github.com/goccy/go-json"
+	"github.com/sirupsen/logrus"
+	"github.com/tbd54566975/ssi-service/pkg/server/framework"
+	"github.com/tbd54566975/ssi-service/pkg/service/common"
+)
+
+type PageToken struct {
+	EncodedQuery  string
+	NextPageToken string
+}
+
+const (
+	PageSizeParam  = "pageSize"
+	PageTokenParam = "pageToken"
+)
+
+func ParsePaginationParams(c *gin.Context, pageRequest *PageRequest) bool {
+	pageSizeStr := framework.GetParam(c, PageSizeParam)
+
+	if pageSizeStr != nil {
+		pageSize, err := strconv.Atoi(*pageSizeStr)
+		if err != nil {
+			errMsg := fmt.Sprintf("list DIDs by method request encountered a problem with the %q query param", PageSizeParam)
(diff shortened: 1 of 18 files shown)


pkg/server/pagination/pagination.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func PageTokenQuery(c *gin.Context) url.Values {
+	query := c.Request.URL.Query()
+	delete(query, PageTokenParam)
+	delete(query, PageSizeParam)
+	return query
+}
+
+func MaybeSetNextPageToken(c *gin.Context, serviceNextPageToken string, respNextPageToken *string) bool {
```
decentralgabe, line 78: would help to add a comment on what this does, esp. why one param is nil-able

pkg/server/pagination/pagination.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+		encodedToken := base64.RawURLEncoding.EncodeToString(nextPageTokenData)
+		*respNextPageToken = encodedToken
+	}
+	return false
+}
+
+type PageRequest struct {
+	// When nil, all DIDs will be returned.
+	PageSize  *int    `json:"pageSize,omitempty"`
```
decentralgabe, line 98: is a page size of 0 valid?

pkg/server/pagination/pagination.go
```diff
@@ -0,0 +1,117 @@
...
+	PageToken *string `json:"pageToken,omitempty"`
```
decentralgabe, line 99: similarly, empty string?

don't think these need to be ptrs

pkg/server/pagination/pagination.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func (r *PageRequest) ToServicePage() common.Page {
```
decentralgabe, line 102: need to be exported?

decentralgabe (approved):

pkg/server/pagination/pagination.go
andresuribe87: Yes, it's used by other packages.

pkg/server/pagination/pagination.go
andresuribe87: Added documentation. It's meant to mimic what was parsed from the request query, so empty means it wasn't provided.

pkg/server/pagination/pagination.go
andresuribe87: It's not. Added validation.

pkg/server/pagination/pagination.go
andresuribe87: Added documentation

PR #539, opened 2023-06-14
Batch create for did:key dids.

andresuribe87:
# Overview
This includes the changes to support batch creation of did:key DIDs. 

# Description
Part of #347 . This PR also changes the endpoint names to be `batch` instead of `batchCreate`. 

# How Has This Been Tested?
Unit and integration tests.


decentralgabe (commented):
what's the use case for the batch creation of custodial DIDs?

andresuribe87:
> what's the use case for the batch creation of custodial DIDs?

An operator of SSI wanting to create DIDs for a list of users they already have.

decentralgabe (approved):

PR #551, opened 2023-06-21
show an example of gin middleware for authorization and authentication

michaelneale:
# Overview
Implementers of an ssi-service may want to weave in authentication and authorization.

# Description
This shows how it can be done with stubbed out code which can be modified as needed. 
There isn't an easy runtime plugin mechanism for golang so this would likely be done by the implementer.

# How Has This Been Tested?

Regression tests covers this as it adds no functionality.

## References
https://github.com/gin-gonic/gin/blob/master/docs/doc.md#using-basicauth-middleware

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,10 @@ Note that at this time, we do not currently support rotating the master key.
 - If does not exist...it uses a default config defined in the code inline
 3. Finally, it loads the config/.env file and adds the env variables defined in this file to the final SSIServiceConfig
 
+### Authentication and Authorization
+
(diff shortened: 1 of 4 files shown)


decentralgabe (changes requested):
I don't think we should add this to the main codeline, I'm fine with it living in an `examples` directory for extension.

I'm not certain we should add any auth, let alone user-based auth like this. If we do add auth I'd favor a DID-auth based solution. Until we have requirements, a design, and some consensus let's hold off on this.

michaelneale:
@decentralgabe yeah can be in examples/docs for sure. I think there needs to be flexibility for implementers. Is there did based auth already in there?

decentralgabe:
no DID auth yet, we have a ticket to track https://github.com/TBD54566975/ssi-sdk/issues/306

decentralgabe (approved):
Lgtm

PR #552, opened 2023-06-21
Distinguish between container.ID and container.Credential.ID

### LINKED ISSUE #544: [Bug] ID includes `localhost`
**Describe the bug**
When using the Credential API, the service returns the `id` field with `http://localhost`

**To Reproduce**
1. Create a credential with the service.
2. Observe the `id` field in the response. It should start with `localhost`
3. Call GET to get credentials from the service.
4. Observe the `id` field in the response. It should start with `localhost`

**Expected behavior**
The `id` field should only return the credential ID

**Supporting Material**
Screenshot 2023-06-15 at 3 22 18 AM

andresuribe87:
# Overview
Fixed #544 by distinguishing between `credentialResponse.ID` and `credentialResponse.Credential.ID`.

# Description

This PR makes explicit the distinction between the two elements, which previously caused confusion. Documentation was also updated. 

# How Has This Been Tested?
All tests were updated with simple renaming.


nitro-neal:
LGTM, just to be clear, 

the top level will be just a guid
the nested cred object will be a localhost:8080/guid ?

nitro-neal (approved):

pkg/service/credential/storage.go
```diff
@@ -21,17 +20,15 @@ import (
...
 }
 
 type StoredCredential struct {
 	// This Key is generated by the storage module upon first write.
 	Key string `json:"key"`
 
-	// This is the same as `credential.id`. Typically, a URL that can be dereferenced.
-	CredentialID string `json:"credentialId"`
+	// ID of the credential that identifies it within ssi service.
+	ServiceID string `json:"serviceId"`
```
decentralgabe, line 31: perhaps something like `localCredentialID` and `externalCredentialID` may be more clear? or internal/external

decentralgabe (approved):

andresuribe87:
> LGTM, just to be clear,
> 
> the top level will be just a guid the nested cred object will be a localhost:8080/guid ?

Correct!

pkg/service/credential/storage.go
andresuribe87: `externalCredentialID` is the same as `Credential.ID`. I don't think we should modify that because it's maps to the spec definition of `id`. 

Renamed this to `localCredentialID`.

Issue #550, opened 2023-06-20
Add support for evidence in VCs


decentralgabe:
Evidence should be added as an optional field https://github.com/TBD54566975/ssi-service/blob/main/pkg/server/router/credential.go#L99

It is supported in the SDK https://github.com/TBD54566975/ssi-sdk/blob/main/credential/model.go#L29

Since it is an any, we may want to pre-populate some fields like `id` and `type` and leave the rest as free form.

This is a prerequsite to uptake in the UI; cc: @kirahsapong

nitro-neal:
pr - https://github.com/TBD54566975/ssi-service/pull/553

PR #553, opened 2023-06-21
Adding optional evidence field to credential creation

nitro-neal:
# Overview
Adding optional evidence field to create credential

# Description
Adding evidence as described here - https://www.w3.org/TR/vc-data-model/#evidence
fixes - https://github.com/TBD54566975/ssi-service/issues/550

```
{
  "@context": [
    "https://www.w3.org/2018/credentials/v1",
    "https://www.w3.org/2018/credentials/examples/v1"
  ],
  "id": "http://example.edu/credentials/3732",
  "type": ["VerifiableCredential", "UniversityDegreeCredential"],
  "issuer": "https://example.edu/issuers/14",
  "issuanceDate": "2010-01-01T19:23:24Z",
  "credentialSubject": {
    "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
    "degree": {
      "type": "BachelorDegree",
      "name": "Bachelor of Science and Arts"
    }
  },
  "evidence": [{
    "id": "https://example.edu/evidence/f2aeec97-fc0d-42bf-8ca7-0548192d4231",
    "type": ["DocumentVerification"],
    "verifier": "https://example.edu/issuers/14",
    "evidenceDocument": "DriversLicense",
    "subjectPresence": "Physical",
    "documentPresence": "Physical",
    "licenseNumber": "123AB4567"
  }],
  "proof": {  }
}
```

# How Has This Been Tested?
* Added to integration test
* Added to unit test


nitro-neal:
Should we add validation code to the sdk?
https://github.com/TBD54566975/ssi-sdk/blob/main/credential/model.go#L29

It seems like the only validation that is required from the spec is:

```
The precise content of each evidence scheme is determined by the specific evidence [type](https://www.w3.org/TR/vc-data-model/#dfn-type) definition.
```

image

pkg/service/credential/model.go
```diff
@@ -24,6 +24,7 @@ type CreateCredentialRequest struct {
 	Expiry      string         `json:"expiry,omitempty"`
 	Revocable   bool           `json:"revocable,omitempty"`
 	Suspendable bool           `json:"suspendable,omitempty"`
+	Evidence    []any          `json:"evidence,omitempty"`
```
decentralgabe, line 27: should we create another type here that always has `id` and `type` fields?

decentralgabe (approved):

pkg/service/credential/model.go
nitro-neal: Yea thats a good way to do it, i'll make the change

doc/swagger.yaml
```diff
@@ -988,6 +988,14 @@ definitions:
         example:
           alumniOf: did_for_uni
         type: object
+      evidence:
+        additionalProperties:
+          type: string
+        description: Optional. Corresponds to `evidence` in https://www.w3.org/TR/vc-data-model/#evidence.
+        example:
+          ' {type': ' [''DocumentVerification''}]'
+          '[{id': ' ''https://example.edu/evidence/f2aeec97-fc0d-42bf-8ca7-0548192d4231''}'
```
andresuribe87, line 997: Not sure if this is rendering correctly?

pkg/server/router/credential.go
```diff
@@ -125,6 +124,9 @@ type CreateCredentialRequest struct {
 	// Whether this credential can be suspended. When true, the created VC will have the "credentialStatus"
 	// property set.
 	Suspendable bool `json:"suspendable,omitempty" example:"false"`
+
+	// Optional. Corresponds to `evidence` in https://www.w3.org/TR/vc-data-model/#evidence.
```
andresuribe87, line 128:
```suggestion
	// Optional. Corresponds to `evidence` in https://www.w3.org/TR/vc-data-model-2.0/#evidence
```

pkg/server/router/credential.go
```diff
@@ -125,6 +124,9 @@ type CreateCredentialRequest struct {
...
+	Evidence []any `json:"evidence" swaggertype:"object,string" example:"[{id: 'https://example.edu/evidence/f2aeec97-fc0d-42bf-8ca7-0548192d4231'}, {type: ['DocumentVerification'}]"`
```
andresuribe87, line 129: I think?
```suggestion
	Evidence []any `json:"evidence" swaggertype:"object,string" example:"[{'id': 'https://example.edu/evidence/f2aeec97-fc0d-42bf-8ca7-0548192d4231'}, {'type': ['DocumentVerification'}]"`
```

pkg/service/credential/model.go
andresuribe87: I would suggest we go even further, and only allow specific types of evidence. That way, we can validate that the input and shape is correct.

andresuribe87 (approved):

nitro-neal:
@andresuribe87  Like only have "type": ["DocumentVerification"], and a few others? 
And then make sure the DocumentVerification only matches the matches the DocumentVerification schema?

Do you know where can I find what DocumentVerification is and what fields it requires?

andresuribe87:
> Like only have "type": ["DocumentVerification"], and a few others?

Similar. To be precise, the only `type` I see us supporting is what's specified in https://docs.google.com/document/d/1y65HA9ZP58ppySG-yiBhGdNwsFrJZ8_-zupOyasuqgk/edit#heading=h.pvfljv28woox (i.e. `https://openid.bitbucket.io/ekyc/openid-connect-4-identity-assurance.html#name-verification-element`). 

> Do you know where can I find what DocumentVerification is and what fields it requires?

AFAIK, no one has implemented the `DocumentVerification` type. We should implement the type described above, since that's going to be used for KYC purposes.

decentralgabe:
I don't think we should implement a specific type until it's in the VC specs directory. I'm in favor of leaving it more generic for now.

andresuribe87:
> I don't think we should implement a specific type until it's in the VC specs directory. I'm in favor of leaving it more generic for now.

SGTM

PR #369, opened 2023-03-31
Implement OIDC4VCI Credential Endpoint

andresuribe87:
# Overview
This change implements https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-credential-endpoint

# Description
In order to support this, some prerequisites were implemented: 
* A middleware called `introspect` that can be set on a per handler basis. It's responsible for gating access to the endpoint defined by the handler.
* Scaffolding for the endpoint.

The implementation itself includes:
* Using a time based one time password for the nonce implementation. The seed key used is the serviceKey already stored in the keyService.
* Models for the request and response.

# How Has This Been Tested?
Multiple unit tests.

# Checklist

Before submitting this PR, please make sure:

- [x] I have read the CONTRIBUTING document.
- [x] My code is consistent with the rest of the project 
- [x] I have tagged the relevant reviewers and/or interested parties
- [x] I have updated the READMEs and other documentation of affected packages

## References
https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-credential-endpoint

PR #556, opened 2023-06-22
ability to set a default bearer token

michaelneale:
# Overview
Allow ability to have Bearer token optionally.

# Description
By setting an AUTH_TOKEN api calls can be checked against a bearer token.

# How Has This Been Tested?
A new test case for default behavior has been added.

## References
https://github.com/zalando/gin-oauth2


andresuribe87 (commented):
I think we should do the introspection approach that was introduced in https://github.com/TBD54566975/ssi-service/pull/369/files (see the `introspect` middleware). We can separate that out into a single PR.

Integrating with an existing authorization server is how I expect production deployments to work.

decentralgabe (commented):
I like this approach but would suggest that we pause on auth until we have a (light) [SIP](https://github.com/TBD54566975/ssi-service/tree/main/sip) in place that goes over options and examples.

michaelneale:
@andresuribe87 yeah I like that - what we want to be careful of is shipping without something default on (as it will come back to bite) and default means assuming something pretty basic (but yeah a rel prod will integrate with many things).

michaelneale:
cc @mistermoe @bradleydwyer if needed can easily add in

decentralgabe:
I believe this should be merged, but since docs have shifted, @michaelneale can you move the README additions to a new piece of [service documentation here](https://github.com/TBD54566975/ssi-service/tree/main/doc#service-documentation) on auth?

michaelneale:
@decentralgabe yep working on it...

pkg/server/server.go
```diff
@@ -118,6 +118,7 @@ func setUpEngine(cfg config.ServerConfig, shutdown chan os.Signal) *gin.Engine {
 		gin.Recovery(),
 		gin.Logger(),
 		middleware.Errors(shutdown),
+		middleware.AuthMiddleware(),
```
decentralgabe, line 121: I think we want to comment this out

decentralgabe (approved):
nice

pkg/server/server.go
michaelneale: it does skip it: https://github.com/TBD54566975/ssi-service/pull/556/files#diff-6320ecdc95f4d46f3bb6598130be227c00be32852a9774c073715363bae0c374R33 if not set

michaelneale:
@decentralgabe ok to merge?

Issue #554, opened 2023-06-21
DID Web creation override


nitro-neal:
**Describe the bug**
When you create 2 did webs with the same id, it 'deletes' the old one

**To Reproduce**
<img width="791" alt="image" src="https://github.com/TBD54566975/ssi-service/assets/5314059/fb6c7128-7df9-4632-9875-f616074b46a9">

<img width="814" alt="image" src="https://github.com/TBD54566975/ssi-service/assets/5314059/ea2618f7-2d4e-4337-88a3-97b03bffe9ce">

<img width="775" alt="image" src="https://github.com/TBD54566975/ssi-service/assets/5314059/cd31650b-d267-4021-a6aa-67484e2c82b4">

<img width="790" alt="image" src="https://github.com/TBD54566975/ssi-service/assets/5314059/5b5ec176-e65a-4873-91c8-e9e73fa2e1c7">

Notice the X is different and the old did web is 'gone?' 

Is this by design ?

nitro-neal:
For this how about we can't make another did - web with the same id?
that should fix the issue

decentralgabe:
this is a bug we should fail if we try to overwrite, also `didWebID` should be changed to `didWebId`

PR #559, opened 2023-06-22
Fixing duplicate did:webs bug

nitro-neal:
# Overview
You can no longer create 2 did webs with the same didWebId

# Description
Fixes https://github.com/TBD54566975/ssi-service/issues/554 

# How Has This Been Tested?
Added to unit test
Manual postman test

image

### DIFF
--- a/pkg/service/did/web.go
+++ b/pkg/service/did/web.go
@@ -36,7 +36,7 @@ var _ MethodHandler = (*webHandler)(nil)
 
 type CreateWebDIDOptions struct {
 	// e.g. did:web:example.com
-	DIDWebId string `json:"didWebId" validate:"required"`
+	DIDWebID string `json:"didWebId" validate:"required"`
 }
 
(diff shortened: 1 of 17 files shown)


pkg/service/did/web.go
```diff
@@ -36,7 +36,7 @@ var _ MethodHandler = (*webHandler)(nil)
 
 type CreateWebDIDOptions struct {
 	// e.g. did:web:example.com
-	DIDWebID string `json:"didWebId" validate:"required"`
+	DIDWebId string `json:"didWebId" validate:"required"`
```
decentralgabe, line 39: oh, this was correct, I'm not sure why your json showed up as `didWebID` maybe either is fine?

I would revert this

decentralgabe (approved):

pkg/service/did/web.go
nitro-neal: yup linter caught this

PR #562, opened 2023-06-23
Support adding/removing services and keys to an ion DID

andresuribe87:
# Overview
Support adding/removing services and keys to an ion DID. Must be custodied.

# Description
This is the last part on #340. Most of the complexity comes from needing to ensure concurrent updates work.

Note that this PR uses the same abstractions from the SDK.

# How Has This Been Tested?
Wrote a unit test.


pkg/service/did/ion.go
```diff
@@ -88,6 +91,287 @@ func (i ionStoredDID) IsSoftDeleted() bool {
...
+	}
+
+	watchKeys := []storage.WatchKey{
+		{
+			Namespace: updateRequestStatusNamespace,
+			Key:       request.DID.String(),
+		},
+	}
+
+	if state.status == "" {
```
decentralgabe, line 140: could this be inverted to reduce nesting?

pkg/service/did/ion.go
```diff
@@ -88,6 +91,287 @@ func (i ionStoredDID) IsSoftDeleted() bool {
...
+			if err != nil {
+				return nil, err
+			}
+
+			updateOp, err := ion.NewUpdateRequest(didSuffix, updateKey, *nextUpdateKey, *signer, request.StateChange)
+			if err != nil {
+				return nil, errors.Wrap(err, "creating update request")
+			}
+
+			state.preAnchor = &preAnchor{
```
decentralgabe, line 164: this should be blocked until 539 is in, no?

pkg/service/did/ion.go
decentralgabe: or consider using a switch and 3 helper methods for readability

pkg/service/did/ion.go
```diff
@@ -88,6 +91,287 @@ func (i ionStoredDID) IsSoftDeleted() bool {
...
+	if state.status == "" {
+		execResp, err := h.storage.db.Execute(ctx, func(ctx context.Context, tx storage.Tx) (any, error) {
+			didSuffix, err := request.DID.Suffix()
+			if err != nil {
+				return nil, errors.Wrap(err, "getting did suffix")
+			}
+
+			updateKey := updatePrivateKey.ToPublicKeyJWK()
+
+			signer, err := ion.NewBTCSignerVerifier(*updatePrivateKey)
```
decentralgabe, line 149: why not call https://github.com/TBD54566975/ssi-sdk/blob/main/did/ion/operations.go#L175 instead

pkg/server/router/did.go
```diff
@@ -127,6 +128,100 @@ func (dr DIDRouter) CreateDIDByMethod(c *gin.Context) {
...
+//	@Failure		500		{string}	string	"Internal server error"
+//	@Router			/v1/dids/{method}/{id} [put]
+func (dr DIDRouter) UpdateDIDByMethod(c *gin.Context) {
+	method := framework.GetParam(c, MethodParam)
+	if method == nil {
+		errMsg := "update DID by method request missing method parameter"
+		framework.LoggingRespondErrMsg(c, errMsg, http.StatusBadRequest)
+		return
+	}
+	if *method != didsdk.IONMethod.String() {
```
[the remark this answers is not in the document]
andresuribe87: The service layer method is called `UpdateIONDID`, to which you pass in an `UpdateIONDIDRequest` object. By design, it's not possible to call that method without an ION did.

The external facing API is open for evolution. The internal is specific so it's more clear and it's easier to maintain.

pkg/server/router/did.go
```diff
@@ -127,6 +128,100 @@ func (dr DIDRouter) CreateDIDByMethod(c *gin.Context) {
...
+	resp := CreateDIDByMethodResponse{DID: createDIDResponse.DID}
+	framework.Respond(c, resp, http.StatusCreated)
+
+}
+
+func toUpdateDIDRequest(id string, request UpdateDIDByMethodRequest) (*did.UpdateDIDRequest, error) {
+	didION := ion.ION(id)
+	if !didION.IsValid() {
+		return nil, errors.Errorf("invalid ion did %s", id)
+	}
```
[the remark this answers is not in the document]
andresuribe87: I added this validation to `ionHandler.UpdateDID`. Could validate in both places, but the `ionHandler` one seems more appropriate in case other endpoints end up calling that method.

pkg/service/did/ion.go
andresuribe87: I don't think it can be inverted. This works like a state machine. When all updates succeed, status will reflect all 3 states within the same call.

When there are failures in the middle, the last state is kept. This is done to ensure that our version of a DID is kept in sync with what was anchored.

pkg/service/did/ion.go
andresuribe87: I wasn't able to find a way in which I could create an instance of `ion.DID` with a custom update key. The only way to create an instance of that struct is by calling [NewIONDID](https://github.com/TBD54566975/ssi-sdk/blob/9617360e9ee224b9ebda63eb52c214bc771e7cfa/did/ion/operations.go#L123), which creates new update keys. 

I agree that it would be desirable, but didn't want to block this PR on an SDK update.

pkg/service/did/ion.go
andresuribe87: Depends on how much pushback there will be for #539 :)

pkg/service/did/service.go
```diff
@@ -167,6 +167,18 @@ func (s *Service) CreateDIDByMethod(ctx context.Context, request CreateDIDReques
 	return handler.CreateDID(ctx, request)
 }
 
+func (s *Service) UpdateIONDID(ctx context.Context, request UpdateDIDRequest) (*UpdateDIDResponse, error) {
```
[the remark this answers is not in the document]
andresuribe87: I do think it's premature. I propose we wait for when we have at least 2 methods for which we'll support updates. 

Currently, it's only did:ion and I don't foresee others in the pipeline either.

pkg/service/did/storage.go
```diff
@@ -69,7 +69,7 @@ func NewDIDStorage(db storage.ServiceStorage) (*Storage, error) {
 	return &Storage{db: db}, nil
 }
 
-func (ds *Storage) StoreDID(ctx context.Context, did StoredDID) error {
+func StoreDID(ctx context.Context, did StoredDID, tx storage.Tx) error {
```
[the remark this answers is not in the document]
andresuribe87: Done

pkg/service/did/ion.go
```diff
@@ -88,6 +91,287 @@ func (i ionStoredDID) IsSoftDeleted() bool {
...
+		}
+	}
+
+	return &UpdateDIDResponse{
+		DID: state.preAnchor.storedDID.DID,
+	}, nil
+}
+
+func isSignatureError(_ error) bool {
+	// TODO: figure out how to determine this error from the body of the response.
```
andresuribe87, line 267: @decentralgabe could you help pointing me to the type of error that I would get if I send an operation to an ION node which has already been applied?

pkg/server/router/did.go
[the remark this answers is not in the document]
decentralgabe: sounds good

pkg/service/did/ion.go
decentralgabe: ok fair

pkg/server/router/did.go
```diff
@@ -127,6 +128,100 @@ func (dr DIDRouter) CreateDIDByMethod(c *gin.Context) {
...
+
+	updateDIDRequest, err := toUpdateIONDIDRequest(*id, request)
+	if err != nil {
+		errMsg := fmt.Sprintf("%s: could not update DID for method<%s>", invalidRequest, *method)
+		framework.LoggingRespondErrWithMsg(c, err, errMsg, http.StatusBadRequest)
+		return
+	}
+	updateIONDIDResponse, err := dr.service.UpdateIONDID(c, *updateDIDRequest)
+	if err != nil {
+		errMsg := fmt.Sprintf("could not create DID for method<%s>", *method)
```
nitro-neal, line 199: is this the right error message?

nitro-neal:
if we remove pub keys / add other pub keys, and the private keys are not in the system, this will be like a "gimp" did.

It wont be able to do things like issue a vc and stuff, I think we have good error messages that would explain that if you try to do something like that.

decentralgabe (approved):
@andresuribe87 can we push this through?

PR #565, opened 2023-06-26
Only use a single encryption key across instances

### LINKED ISSUE #560: [Bug] ServiceKey is overwritten on every restart when MasterKeyPassword is used
**Describe the bug**
We generate a [ServiceKey](https://github.com/TBD54566975/ssi-service/blob/main/pkg/service/keystore/storage.go#L108) on every call to [NewEncryption](https://github.com/TBD54566975/ssi-service/blob/main/pkg/service/keystore/storage.go#L99). The result is that encryption is broken across deployments/pods.

**To Reproduce**
1. Run ssi-service with more than one replica. 
2. Create a key with replica 1. 
3. Read the key with replica 2.
4. Note that the previous step fails with an error.

**Expected behavior**
There should be no error.

andresuribe87:
# Overview
Fixes #560, where multiple instances of the service used different encryption keys.

# Description
This is fixing the bug. 

# How Has This Been Tested?
I tested this by restarting the ssi-service docker container, and checking that the serviceKey didn't change.

### DIFF
--- a/pkg/service/keystore/storage.go
+++ b/pkg/service/keystore/storage.go
@@ -104,27 +104,56 @@ const (
 	awsKMSScheme = "aws-kms"
 )
 
-func NewEncryption(db storage.ServiceStorage, tx storage.Tx, cfg config.KeyStoreServiceConfig) (Encrypter, Decrypter, error) {
+// EnsureServiceKeyExists makes sure that the service key that will be used for encryption exists. This function is
+// idempotent, so that multiple instances of ssi-service can call it on boot.
+func EnsureServiceKeyExists(config config.KeyStoreServiceConfig, provider storage.ServiceStorage) error {
+	if config.MasterKeyURI != "" {
+		return nil
+	}
+
+	watchKeys := []storage.WatchKey{{
+		Namespace: namespace,
+		Key:       skKey,
+	}}
+
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
 
+	_, err := provider.Execute(ctx, func(ctx context.Context, tx storage.Tx) (any, error) {
+		// Create the key only if it doesn't already exist.
+		gotKey, err := getServiceKey(ctx, provider)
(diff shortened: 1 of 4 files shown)


pkg/service/keystore/storage.go
```diff
@@ -104,27 +104,56 @@ const (
 	awsKMSScheme = "aws-kms"
 )
 
-func NewEncryption(db storage.ServiceStorage, tx storage.Tx, cfg config.KeyStoreServiceConfig) (Encrypter, Decrypter, error) {
+// EnsureServiceKeyExists makes sure that the service key that will be used for encryption exists. This function is
+// idempotent, so that multiple instances of ssi-service can call it on boot.
+func EnsureServiceKeyExists(config config.KeyStoreServiceConfig, provider storage.ServiceStorage) error {
```
decentralgabe, line 109: does this need to be exported?

pkg/service/keystore/storage.go
```diff
@@ -104,27 +104,56 @@ const (
...
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
 
-	if len(cfg.MasterKeyURI) != 0 {
-		return NewExternalEncrypter(ctx, cfg)
-	}
-
-	// First, generate a service key
-	serviceKey, serviceKeySalt, err := GenerateServiceKey(cfg.MasterKeyPassword)
+	_, err := provider.Execute(ctx, func(ctx context.Context, tx storage.Tx) (any, error) {
```
decentralgabe, line 122: in this function is `any` used? I only see nil returned

pkg/service/keystore/storage.go
andresuribe87: Yes. It is called during the instantiation of SSi services, before creating the keystore service factory.

pkg/service/keystore/storage.go
andresuribe87: It is not. The signature still needs it because the execute method expects one of its parameters to be a function that implements a specific signature.

decentralgabe (approved):

PR #568, opened 2023-06-28
makes expiration optional for request endpoints

### LINKED ISSUE #557: Passing no expiration to v1/presentations/request should not set a default expiration
There has been discussion of why the current behavior is to set a default expiration on the presentation request that's being created. It's proposed to not set expiration in the presentation request JWT.

See https://discord.com/channels/937858703112155166/969272692891086868/1121290626999263333 and the subsequent messages. 

The summary for reasons to adopt not setting the `exp` claim in the JWT: 
* The use of `exp` is optional in the JWT spec.
* When the presentation request is meant to be static and not change, it doesn't make sense to expire it.
* Imposing expiration as a requirement ends up becoming a burden.

The summary for reasons for always setting the `exp` claim in the JWT that represents the request.
* It acts as a mitigation in case that the presentation request needs to be invalidated.
  * This is not preventing an attack vector, but it's mitigation if an attack happens (I had communicated this incorrectly in the past).
* We don't expose expiration functionality of requests in the server side.
* Presentation requests will support replay protection in the future (this is NOT supported right now). The current thinking is that replay protection should be enabled by default. For single use presentation requests, including the expiration date is a good practice. An example of this is the NIST recommendation for address confirmation. See Section 4.4.1.6, item 5.e of https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-63a.pdf. The equivalent of enrollment codes are `nonce`. 

My personal take
* While I believe that "presentation request is meant to be static and not change" will end up actually changing, I think we don't know yet what a "best practice" will be related to Presentation Requests. So I think we should meet the current needs and figure it out from there. 
* Additionally, we can make it a RECOMMENDATION, instead of a MUST.

andresuribe87:
# Overview
Fixes #557. This makes expiration optional for any requests that are created. 

# Description
Changes a default behavior according to the issue.

# How Has This Been Tested?
Unit and integration tests pass.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -171,7 +171,6 @@ func (o *OperationServiceConfig) IsEmpty() bool {
 
 type PresentationServiceConfig struct {
 	*BaseServiceConfig
-	ExpirationDuration time.Duration `toml:"expiration_duration" conf:"default:30m"`
 }
 
 func (p *PresentationServiceConfig) IsEmpty() bool {
@@ -183,7 +182,6 @@ func (p *PresentationServiceConfig) IsEmpty() bool {
 
 type ManifestServiceConfig struct {
 	*BaseServiceConfig
-	ExpirationDuration time.Duration `toml:"expiration_duration" conf:"default:30m"`
 }
 
 func (m *ManifestServiceConfig) IsEmpty() bool {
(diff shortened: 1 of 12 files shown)


decentralgabe (approved):
I still have reservations about this change, since we're giving users of the service a foot gun, and they might not be aware of it--specifically around having a request they no longer accept responses against.

That said, I'm ok merging as is given the contention.

PR #570, opened 2023-06-29
Ability to provide public keys when creating an ION did

andresuribe87:
# Overview
Upon creation of an ion did, clients can now provide a public key to associate with the did, for which they own the private keys.

# Description
Some quality of life improvements were algo included in this PR. A follow up will change the API surface so that it's clear what `issuerKID` means and we avoid confusion. 

Separately, instead of constructing the did doc, we're taking the one returned from the ION node. 

# How Has This Been Tested?
Integration and unit

### DIFF
--- a/pkg/service/did/ion.go
+++ b/pkg/service/did/ion.go
@@ -9,6 +9,7 @@ import (
 	"github.com/TBD54566975/ssi-sdk/crypto/jwx"
 	"github.com/TBD54566975/ssi-sdk/did"
 	"github.com/TBD54566975/ssi-sdk/did/ion"
+	"github.com/TBD54566975/ssi-sdk/did/resolution"
 	"github.com/TBD54566975/ssi-sdk/util"
 	"github.com/google/uuid"
 	"github.com/mr-tron/base58"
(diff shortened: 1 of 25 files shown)


pkg/service/did/ion.go
```diff
@@ -120,6 +117,7 @@ func (h *ionHandler) CreateDID(ctx context.Context, request CreateDIDRequest) (*
 			Purposes: []ion.PublicKeyPurpose{ion.Authentication, ion.AssertionMethod},
 		},
 	}
+	pubKeys = append(pubKeys, opts.PublicKeys...)
```
decentralgabe, line 120: I'm not sure we should allow adding any public keys without a signature. I'd recommend changing it to a `SignedPublicKeyRequest` of some sort.

andresuribe87: > I'm not sure we should allow adding any public keys without a signature.

Would you mind explaining the rationale? Is the intention here to ensure that the requester has control of the private key associated with the public key that's being added?

If so, is it ok to add this as a separate feature? There are details I think would need to be ironed out before doing so including: should the signature be of a hardcoded value or of a nonce? If the latter, how will the nonce be generated and communicated?

decentralgabe: > Would you mind explaining the rationale? Is the intention here to ensure that the requester has control of the private key associated with the public key that's being added?

Yes exactly.

> If so, is it ok to add this as a separate feature? 

Sure, let's open an issue.

decentralgabe (approved):

Issue #525, opened 2023-06-08
Protect against running out of memory


andresuribe87:
In the current redis.go there is still this comment:
// TODO: This potentially could dangerous as it might run out of memory as we populate allKeys

Is this not true anymore right with the changes to 
func readAllKeys(ctx context.Context, namespace string, b *RedisDB, pageSize int, cursor uint64) ([]string, string, error) {
?

_Originally posted by @nitro-neal in https://github.com/TBD54566975/ssi-service/pull/518#discussion_r1223270812_
            

We should remove all calls that set pageSize == -1 internally. Furthermore, we should iterate through the results in a memory friendly way, and set configurable limits of the pageSize.

radurobot:
I've created a pull request that may resolve this issue. Here's the link to the PR: https://github.com/TBD54566975/ssi-service/pull/575. I've revised the `readAll` function to use a more memory-efficient approach when retrieving keys from the Redis database. Please review the changes and let me know if there are any concerns.

PR #575, opened 2023-07-05
address potential memory issues in Redis readAll function

radurobot:
# Overview
This PR addresses a potential memory issue in the `readAll` function. The function was originally designed to retrieve all keys at once, which could potentially lead to out-of-memory errors for large numbers of keys. Issue: #525 

# Description
The `readAll` function has been revised to use a goroutine and a channel to process each key-value pair individually, which should mitigate the risk of running out of memory. The function ranges over the channel to process each key-value pair and add it to the result map. This approach should be more memory-efficient than the original function, while still only making a single round-trip call to the Redis server.

# How Has This Been Tested?
The function's output and overall logic remain the same, and it has been tested to ensure it still works as expected. Further testing and review are welcome.

# Checklist

- [X] My code is consistent with the rest of the project 
- [X] I have tagged the relevant reviewers and/or interested parties
- [X] I have updated the READMEs and other documentation of affected packages


decentralgabe (approved):
thanks for your contribution! 

will wait for @andresuribe87 to merge

andresuribe87:
Thanks @radurobot for your contribution!

 #525 refers to [this line](https://github.com/TBD54566975/ssi-service/blob/c8541a8597ab1cd8ad1d650d7d682f58e5119ad5/pkg/storage/redis.go#L306). Upon first review, I'm unsure how this PR would address that. Could you clarify what would happen if the DB returns more objects that would not fit in the physical memory of the ssi-service instance?

andresuribe87:
@radurobot that's an awesome solution! But I worry it might just be overcomplicating the problem. I was thinking of something simpler, like returning an error when `readAll` is called with more than 1 Million keys (or similar).

For the full fledged solution, this would involve making sure the most List* endpoints are supporting pagination.

andresuribe87:
> What do you think?

Pagination on List* endpoint is my preferred way to move forward. Some of the benefits include:
* It's a well known pattern. 
* We don't run the risk of OOM when the the collection has too many element.
* It prevents developers from shooting themselves in the foot accidentally, by calling `readAll` on collections that are more than a predefined amount.

Issue #571, opened 2023-06-30
Validate that requester has control of private key for public key passed in creation of ion did


andresuribe87:
> Would you mind explaining the rationale? Is the intention here to ensure that the requester has control of the private key associated with the public key that's being added?

Yes exactly.

> If so, is it ok to add this as a separate feature? 

Sure, let's open an issue.

_Originally posted by @decentralgabe in https://github.com/TBD54566975/ssi-service/pull/570#discussion_r1248009158_

decentralgabe:
We would need a request-response model, where the user makes a request to add a key, we surface a nonce, and they sign over the nonce. The mechanism you describe can work, but needs the nonce - otherwise we risk replay attacks. It's not a huge threat since we're defining a custom format, but I'd argue it's better to be safe.

andresuribe87:
Can you elaborate what the attack is stealing in this case?

The original intent of this issue is to ensure that the client has control of the private key. authn/authz of the did/ion creation endpoint are separate. So it isn't a securing mechanism, but rather it's a way to ensure that the client is passing in the correct key, and that `ssi-service` is able to understand how to validate it.

decentralgabe:
You'd want to make sure that someone couldn't discover a signature proving they control a key and use it for another purpose than they intended, I think what you suggested is fine

PR #577, opened 2023-07-07
`jwsPublicKeys` as an option when creating ION DIDs.

andresuribe87:
# Overview
Fixes https://github.com/TBD54566975/ssi-service/issues/571. 

# Description
See original issue for motivation. This PR add the new field in a backwards compatible way. Documentation has been updated as well to include many details. 

# How Has This Been Tested?
Unit tests and integration tests.

### DIFF
--- a/pkg/service/did/ion.go
+++ b/pkg/service/did/ion.go
@@ -11,7 +11,9 @@ import (
 	"github.com/TBD54566975/ssi-sdk/did/ion"
 	"github.com/TBD54566975/ssi-sdk/did/resolution"
 	"github.com/TBD54566975/ssi-sdk/util"
+	"github.com/goccy/go-json"
 	"github.com/google/uuid"
+	"github.com/lestrrat-go/jwx/v2/jws"
 	"github.com/mr-tron/base58"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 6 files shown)


pkg/service/did/ion.go
```diff
@@ -52,8 +54,24 @@ type ionHandler struct {
...
+	// Services to add to the DID document that will be created.
+	ServiceEndpoints []did.Service `json:"serviceEndpoints"`
+
+	// List of public keys to add to the DID document. Each element must follow the schema described in step 3 of https://identity.foundation/sidetree/spec/#add-public-keys.
+	// Each element will be used to add public keys to the DID document in the same way in which the `add-public-keys`
+	// patch action adds keys (see https://identity.foundation/sidetree/spec/#add-public-keys).
+	//
+	// It is recommended to us `jwsPublicKeys` instead of this method, which guarantees that the caller has control
+	// of the private key associated with the public keys that's being added.
+	PublicKeys []ion.PublicKey `json:"publicKeys"`
```
decentralgabe, line 66: do we need this if the pk is contained in the JWS?

what if the two conflict? we're needing to check for missing / mismatched keys

decentralgabe (commented):
I believe we should remove the non-authenticated public keys

pkg/service/did/ion.go
andresuribe87: Removed.

decentralgabe (approved):

PR #584, opened 2023-07-10
Separates service internal namespace from externally added keys.

### LINKED ISSUE #367: Decouple ServiceKey namespace from KeyStoreAPI namespace
We are storing the ServiceKey in the same namespace as the one we use for the `KeyStoreAPI`. What this means is that a caller may override the ServiceKey that was stored, if they pass in `id == "ssi-service-key"` in the [Store Key Request](https://developer.tbd.website/docs/apis/ssi-service#tag/KeyStoreAPI/paths/~1v1~1keys/put)

andresuribe87:
# Overview
Fixes #367 by changing the namespace that the service key uses.

# Description
na

# How Has This Been Tested?
All tests pass.

### DIFF
--- a/pkg/service/keystore/storage.go
+++ b/pkg/service/keystore/storage.go
@@ -55,10 +55,15 @@ type ServiceKey struct {
 
 const (
 	namespace             = "keystore"
-	serviceNamespace      = "keystore:service-internal"
-	publicNamespaceSuffix = ":public-keys"
+	serviceInternalSuffix = "service-internal"
+	publicNamespaceSuffix = "public-keys"
 	skKey                 = "ssi-service-key"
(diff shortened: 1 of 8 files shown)


nitro-neal (approved):

pkg/service/keystore/storage.go
```diff
@@ -55,6 +55,7 @@ type ServiceKey struct {
 
 const (
 	namespace             = "keystore"
+	serviceNamespace      = "keystore:service-internal"
```
decentralgabe, line 58: I thought we were using `-` instead of `:` in other places, may want to make sure this is uniform

decentralgabe (approved):

PR #585, opened 2023-07-11
Making applicant required

nitro-neal:
image


nitro-neal:
Note: Any downstream process that is constructing an applicationJWT without the applicant field will fail now

decentralgabe (approved):

PR #586, opened 2023-07-11
Added DID configuration resource creation endpoint

### LINKED ISSUE #355: .well_known config API endpoint
currently the wellknown config generation is unit test, need to make this an endpoint

andresuribe87:
# Overview
Fixes #355 by adding an API endpoint.

# Description
The endpoint includes instructions for what to do after it is called. It is deliberate that this endpoint does not store the actual configuration resource, as it's meant to be hosted elsewhere. An additional endpoint will be added that can be called to verify a DID Configuration Resource given an origin.

# How Has This Been Tested?
Unit and integration tests


pkg/server/router/did_configuration.go
```diff
@@ -0,0 +1,113 @@
...
+		ExpirationDate:       r.ExpirationDate,
+		IssuanceDate:         r.IssuanceDate,
+	}
+}
+
+type DIDConfiguration struct {
+	Context    any   `json:"@context" validate:"required"`
+	LinkedDIDs []any `json:"linked_dids" validate:"required"`
+}
+type CreateDIDConfigurationResponse struct {
```
decentralgabe, line 57:
```suggestion
	LinkedDIDs []any `json:"linkedDids" validate:"required"`
}

type CreateDIDConfigurationResponse struct {
```

PR #590, opened 2023-07-12
Supporting SQL based storage

andresuribe87:
# Overview
This is a super naive, but working, implementation of how a SQL storage can be implemented.

# Description
This implementations leverages `database/sql`. This means that you can use any SQL db, as long as there is an official driver supported (see https://github.com/golang/go/wiki/SQLDrivers).  

In this PR, I also included some small fixes to make sure that we're using the same `*testing.T` object as the rest of the tests for the `ServiceStorage`. 

# How Has This Been Tested?
I've added unit tests for the new implementation that uses an embedded postgres db. This isn't my favorite because it's slow and sluggish. 

I did not add it as part of unit tests across all the test suite because:
1. There is no way to run the embeddeddb in parallel, which causes all sorts of concurrency problems. This could be fixed by blocking all test, but then it means that all packages need to be run sequentially. 
2. The amount of tests that need to pass balloon.

## References
If you haven't seen how `database/sql` works, look at https://go.dev/doc/tutorial/database-access


decentralgabe (approved):
this is a great proof of concept!

would you mind adding a brief [doc in here](https://github.com/TBD54566975/ssi-service/tree/main/doc) about storage?

PR #592, opened 2023-07-13
Added DID Configuration Verification Endpoint

andresuribe87:
# Overview
This PR adds the `v1/did-configurations/verification` endpoint, with documentation. It follows the spec in https://identity.foundation/.well-known/resources/did-configuration/#did-configuration-resource-verification

# Description
Now that we have an endpoint that can create a DIDConfiguration, this PR adds another endpoint so that clients can verify they've done the correct steps when hosting it (or that any other DID Configuration is correct).

# How Has This Been Tested?
Unit tests were added. Integration tests were not added, since verifying `https://www.tbd.website` fails with the following reason : `verifying JWT credential: error getting key to verify credential<>: did<did:key:z6Mkh4A6QTNCAFi4NZfnWt8qNSGXDGnNXhxXWewcpBrzS19v> has no verification methods with kid: did:key:z6Mkh4A6QTNCAFi4NZfnWt8qNSGXDGnNXhxXWewcpBrzS19v`. I.e. we need to update the JWT to use the verification method ID, instead of the DID. This is tracked in a separate issue. 

# Checklist

## References
https://identity.foundation/.well-known/resources/did-configuration/#did-configuration-resource-verification

### DIFF
--- a/pkg/service/well-known/did_configuration.go
+++ b/pkg/service/well-known/did_configuration.go
@@ -1,24 +1,47 @@
 package wellknown
 
 import (
+	"bytes"
 	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
 	"time"
 
 	"github.com/TBD54566975/ssi-sdk/credential"
 	"github.com/TBD54566975/ssi-sdk/credential/integrity"
 	"github.com/TBD54566975/ssi-sdk/did"
+	"github.com/TBD54566975/ssi-sdk/did/resolution"
+	"github.com/goccy/go-json"
 	"github.com/pkg/errors"
 	credint "github.com/tbd54566975/ssi-service/internal/credential"
+	"github.com/tbd54566975/ssi-service/internal/util"
 	svcframework "github.com/tbd54566975/ssi-service/pkg/service/framework"
 	"github.com/tbd54566975/ssi-service/pkg/service/keystore"
+	"github.com/tbd54566975/ssi-service/pkg/service/schema"
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
 )
 
 type DIDConfigurationService struct {
 	keyStoreService *keystore.Service
+	validator       *credint.Validator
+
+	HTTPClient *http.Client
 }
 
-func NewDIDConfigurationService(keyStoreService *keystore.Service) *DIDConfigurationService {
-	return &DIDConfigurationService{keyStoreService: keyStoreService}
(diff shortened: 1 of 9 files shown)


pkg/service/well-known/did_configuration.go
```diff
@@ -33,6 +56,80 @@ func (s DIDConfigurationService) Status() svcframework.Status {
...
+
+	response := VerifyDIDConfigurationResponse{
+		DIDConfiguration: bodyCopy.String(),
+	}
+	for _, domainLinkageCredential := range didConfiguration.LinkedDIDs {
+		// For a Domain Linkage Credential to be deemed valid, it MUST be successfully processed in accordance with the following steps:
+		//
+		// 1. The credentialSubject.id MUST be a DID, and the value MUST be equal to both the Subject and Issuer of the Domain Linkage Credential.
+		credentialSubjectID := domainLinkageCredential.Credential.CredentialSubject["id"].(string)
+		if !strings.HasPrefix(credentialSubjectID, "did:") {
```
decentralgabe, line 97: believe there's a method `IsValidDID` or similar in the utils somewhere

pkg/service/well-known/did_configuration.go
```diff
@@ -1,24 +1,47 @@
...
 	keyStoreService *keystore.Service
+	validator       *credint.Validator
+
+	HTTPClient *http.Client
 }
 
-func NewDIDConfigurationService(keyStoreService *keystore.Service) *DIDConfigurationService {
-	return &DIDConfigurationService{keyStoreService: keyStoreService}
+func NewDIDConfigurationService(keyStoreService *keystore.Service, didResolver resolution.Resolver, schema *schema.Service) (*DIDConfigurationService, error) {
+	client := &http.Client{Transport: otelhttp.NewTransport(http.DefaultTransport)}
```
decentralgabe, line 34: should this be the default client (with timeouts, etc) and then add in the otel?

pkg/service/well-known/did_configuration.go
andresuribe87: I searched for something similar, but couldn't find anything.

pkg/service/well-known/did_configuration.go
andresuribe87: DefaultClient is the same as the zero value of `http.Client`. Timeouts are carried by the `ctx` when the call is made. 

We _could_ have an `httpClientProvider`, but then that get's into the dependency injection and I wanted to keep this simple.

PR #595, opened 2023-07-19
Add action to notify SSI console on pushes to main

kirahsapong:
This PR adds an action to open an issue in the the SSI Console repo when a change goes to main


decentralgabe (approved):

.github/workflows/release-notifier.yml
```diff
@@ -31,3 +31,31 @@ jobs:
...
+
+jobs: 
+  create-issue:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Notify SSI Console
+      uses: actions/github-script@v6
+      with:
+        github-token: ${{ secrets.DOCS_PAT }}  # Assuming 'PAT' is your Personal Access Token stored as a secret
```
decentralgabe, line 49: @kirahsapong who manages this PAT? is there a way to do it without one?

PR #599, opened 2023-07-25
Added app level encryption feature.

andresuribe87:
# Overview
Allow for operators of ssi-service to choose whether to do app level encryption.

# Description

At a high-level, this PR is adding a Data Encryption Key. The DEK can be stored either in the configured storage, or in a KMS. The DEK is used for every read/write that happens in the application.

Specifically, this PR does the following:
* Removed some methods that weren't necessary in the storage interface. 
* Implemented and `EncryptedWrapper`, which takes in any implementation of a storage, and adds an encryption layer. 
* Refactored the existing Key Encryption Mechanism so it could be reused with the data encryption layer. 
* Added documentation.

# How Has This Been Tested?
All unit tests and integration tests pass when encryption is enabled (see the dev.toml file).


config/config.go
```diff
@@ -94,21 +98,53 @@ type BaseServiceConfig struct {
...
 	*BaseServiceConfig
 
-	// The URI for the master key. We use tink for envelope encryption as described in https://github.com/google/tink/blob/9bc2667963e20eb42611b7581e570f0dddf65a2b/docs/KEY-MANAGEMENT.md#key-management-with-tink
-	// When left empty, then a random key is generated and used.
+	// Configuration describing the encryption of the private keys that are under ssi-service's custody.
+	EncryptionConfig
+}
+
+type EncryptionConfig struct {
+	DisableEncryption bool `toml:"disable_encryption"`
```
decentralgabe, line 106: do we have some circuit-breaker logic that doesn't allow this to be a set, or at least throws a warning if env == prod?

config/config.go
```diff
@@ -72,6 +72,10 @@ type ServicesConfig struct {
 	StorageOptions  []storage.Option `toml:"storage_option"`
```
decentralgabe, line 72: general comment: it would be great to have a light design doc around this kind of thing before going for an implementation. there are a number of hairy details that would be good to discuss and get feedback on. not only that it would serve as a record for how the feature was thought through which will be useful in the future (as we have already seen with existing SIPs)

doc/STORAGE.md
```diff
@@ -70,4 +70,45 @@ For a working example, see this [dev.toml file](https://github.com/TBD54566975/s
...
+disable_encryption = false
+```
+
+Disabling app level encryption is also possible using the following options in your TOML configuration:
+
+```toml
+[services.storage_encryption]
+# encryption
+disable_encryption = true
+```
```
decentralgabe, line 114: this is great to have

pkg/encryption/encryption.go
```diff
@@ -0,0 +1,164 @@
+package encryption
+
+import (
+	"context"
+	"strings"
+
+	util2 "github.com/TBD54566975/ssi-sdk/util"
```
decentralgabe, line 7:
```suggestion
	sdkutil "github.com/TBD54566975/ssi-sdk/util"
```

pkg/encryption/encryption.go
```diff
@@ -0,0 +1,164 @@
...
+// Decrypter is the interface for any decrypter. May be AEAD or Hybrid.
+type Decrypter interface {
+	// Decrypt decrypts ciphertext. The second parameter may be treated as associated data for AEAD (as abstracted in
+	// https://datatracker.ietf.org/doc/html/rfc5116), or as contextInfofor HPKE (https://www.rfc-editor.org/rfc/rfc9180.html)
+	Decrypt(ctx context.Context, ciphertext, contextInfo []byte) ([]byte, error)
+}
+
+type KeyResolver func(ctx context.Context) ([]byte, error)
+
+type XChaCha20Poly1305Encrypter struct {
```
decentralgabe, line 33: is this only used locally since the doc above mentions only AES GCM is supported by HSMs?

pkg/service/keystore/service.go
```diff
@@ -50,20 +49,17 @@ func (s Service) Config() config.KeyStoreServiceConfig {
 }
 
 func NewKeyStoreService(config config.KeyStoreServiceConfig, s storage.ServiceStorage) (*Service, error) {
-	if err := EnsureServiceKeyExists(config, s); err != nil {
-		return nil, sdkutil.LoggingErrorMsg(err, "initializing keystore")
+	encrypter, decrypter, err := NewServiceEncryption(s, config.EncryptionConfig, ServiceKeyEncryptionKey)
```
decentralgabe, line 52: I would prefer, that instead of a `noopEncrypter` and `noopDecrypter` we check config here, and if the flag is off we pass in nil values below, this avoids the overhead of a no-op call for each read/write

pkg/service/keystore/service_test.go
```diff
@@ -23,71 +22,6 @@ func TestGenerateServiceKey(t *testing.T) {
...
-			kt: crypto.P384,
-		},
-		{
-			kt: crypto.P521,
-		},
-		{
-			kt: crypto.RSA,
-		},
-	}
-	for _, test := range tests {
```
decentralgabe, line 61: why delete this?

pkg/service/keystore/storage.go
```diff
@@ -89,45 +80,24 @@ func NewKeyStoreStorage(db storage.ServiceStorage, e Encrypter, d Decrypter, wri
...
-	if config.MasterKeyURI != "" {
+func ensureEncryptionKeyExists(config encryption.ExternalEncryptionConfig, provider storage.ServiceStorage, namespace, encryptionKeyKey string) error {
+	if config.GetMasterKeyURI() != "" {
 		return nil
 	}
 
 	watchKeys := []storage.WatchKey{{
 		Namespace: namespace,
-		Key:       skKey,
+		Key:       encryptionKeyKey,
```
decentralgabe, line 92: maybe `encryptionKeyID`?

pkg/service/service.go
```diff
@@ -81,20 +82,27 @@ func validateServiceConfig(config config.ServicesConfig) error {
...
+	unencryptedStorageProvider, err := storage.NewStorage(storage.Type(config.StorageProvider), config.StorageOptions...)
 	if err != nil {
 		return nil, sdkutil.LoggingErrorMsgf(err, "could not instantiate storage provider: %s", config.StorageProvider)
 	}
 
+	storageEncrypter, storageDecrypter, err := keystore.NewServiceEncryption(unencryptedStorageProvider, config.AppLevelEncryptionConfiguration, keystore.ServiceDataEncryptionKey)
+	if err != nil {
+		return nil, errors.Wrap(err, "creating app level encrypter")
+	}
+	storageProvider := storage.NewEncryptedWrapper(unencryptedStorageProvider, storageEncrypter, storageDecrypter)
```
decentralgabe, line 94: if this is a no-op the encrypted wrapper is unnecessary

pkg/service/service.go
```diff
@@ -81,20 +82,27 @@ func validateServiceConfig(config config.ServicesConfig) error {
...
 	}
 
-	if err := keystore.EnsureServiceKeyExists(config.KeyStoreConfig, storageProvider); err != nil {
-		return nil, sdkutil.LoggingErrorMsg(err, "could not ensure the service key exists")
+	keyEncrypter, keyDecrypter, err := keystore.NewServiceEncryption(unencryptedStorageProvider, config.KeyStoreConfig.EncryptionConfig, keystore.ServiceKeyEncryptionKey)
+	if err != nil {
+		return nil, errors.Wrap(err, "creating keystore encrypter")
 	}
-	keyStoreServiceFactory := keystore.NewKeyStoreServiceFactory(config.KeyStoreConfig, storageProvider)
+	keyStoreServiceFactory := keystore.NewKeyStoreServiceFactory(config.KeyStoreConfig, storageProvider, keyEncrypter, keyDecrypter)
```
decentralgabe, line 105: same comment here

decentralgabe (changes requested):
it would be great to have tests that exercise different confidence levels (no encryption, local encryption, hsm encryption)

for hsm we would need a reliable cloud test env, so maybe that's out of scope for now but would be good to track in a future issue.

similarly, we should track issues for managing key rotation.

if you didn't (I may have missed) it would be useful to highlight that we've specifically chosen an authenticated encryption scheme which gives us the added benefit of tamper evidence, and may also prove useful for decrypting/re-encrypting data during key rotations since we'll be able to see which key encrypted the data


another thing worth covering in the doc is the data that we need to keep in plaintext, and the considerations around that, for keys. these keys are often just UUIDs but sometimes contain other information to make prefix queries viable. that means they could contain a DID, which by someones definition may be PII. <-- not sure here but we should verify

andresuribe87:
> it would be great to have tests that exercise different confidence levels (no encryption, local encryption, hsm encryption)
> 
> for hsm we would need a reliable cloud test env, so maybe that's out of scope for now but would be good to track in a future issue.

no encryption is covered in unit tests. Local encryption is covered in integration tests. HSM will be tracked in https://github.com/TBD54566975/ssi-service/issues/600
> 
> similarly, we should track issues for managing key rotation.

https://github.com/TBD54566975/ssi-service/issues/601

> 
> if you didn't (I may have missed) it would be useful to highlight that we've specifically chosen an authenticated encryption scheme which gives us the added benefit of tamper evidence, and may also prove useful for decrypting/re-encrypting data during key rotations since we'll be able to see which key encrypted the data

I'm going to leave that for the same issue as above (i.e. #601 )

> 
> another thing worth covering in the doc is the data that we need to keep in plaintext, and the considerations around that, for keys. these keys are often just UUIDs but sometimes contain other information to make prefix queries viable. that means they could contain a DID, which by someones definition may be PII. <-- not sure here but we should verify

We _could_ encrypt keys as long as we preserve the key prefix structure. That said, it doesn't seem like it's worth the effort and additional complexity. I added some more info to the docs, and created another issue to track #603

decentralgabe (approved):

andresuribe87 (commented):
Seems like these never went through :O

Issue #606, opened 2023-07-28
Add basic "how to's" for the service's functionality


decentralgabe:
How to's that demonstrate basic usage of the APIs/functionality for:

- [x] Creating a DID - done in https://github.com/TBD54566975/ssi-service/pull/607
- [x] Creating a Schema - done in https://github.com/TBD54566975/ssi-service/pull/610
- [x] Issuing a Credential - done in https://github.com/TBD54566975/ssi-service/pull/610
- [x] Verifying a Credential - done in https://github.com/TBD54566975/ssi-service/pull/616
- [x] Changing the Status of a Credential - done in https://github.com/TBD54566975/ssi-service/pull/616
- [x] Creating a Well Known File for your DID - done in https://github.com/TBD54566975/ssi-service/pull/611
- [ ] #622
- [ ] #623

Maybe some more around webhooks, operations, etc.

decentralgabe:
closing in favor of #622 and #623 initial set of how tos are already done

Issue #502, opened 2023-05-31
Config should not expose non-configurable values


decentralgabe:
this needs to be refactored because we don't want to enable config for setting service endpoints. I'll open a new issue

_Originally posted by @decentralgabe in https://github.com/TBD54566975/ssi-service/pull/494#discussion_r1212022099_

decentralgabe:
Users should not be able to override the endpoints of the REST API for different services.

Consider moving this into a non-configurable object.

decentralgabe:
Similarly, we should not expose setting name, description, and version properties in the config.

PR #607, opened 2023-07-28
Update READMEs, docs, add how-tos section, first how-to for DIDs

decentralgabe:
Address part of #606


michaelneale:
SMASH THE LIKE BUTTON SO HARD I BROKE MY KEYBOARD 
image

michaelneale:
@decentralgabe is this a WIP? If so, can we have a flow on the REAME with curl which issues a credential for the simplest schema? (as one block example) - that would be wonderful

doc/README.md
```diff
@@ -1,11 +1,57 @@
...
+| [Creating a Schema](https://github.com/TBD54566975/ssi-service/blob/main/doc/howto/schema.md)                                                | Get started with schema functionality                  |
+| [Issuing a Credential](https://github.com/TBD54566975/ssi-service/blob/main/doc/howto/credential.md)                                         | Get started with credential issuance functionality     |
+| [Verify a Credential](https://github.com/TBD54566975/ssi-service/blob/main/doc/howto/verification.md)                                        | Get started with credential verification functionality |
+| [Revoke/Suspend a Credential](https://github.com/TBD54566975/ssi-service/blob/main/doc/howto/status.md)                                      | Get started with credential status functionality       |
+| [[TODO] Requesting and Verifying Credentials with Presentation Exchange](https://github.com/TBD54566975/ssi-service/issues/606)              | Get started with Presentation Exchange functionality   |
```
andresuribe87, line 53: Good doc here: https://hackmd.io/kvCAWD79SFKneGabpdmkgQ courtesy of @tomdaffurn

doc/howto/did.md
```diff
@@ -0,0 +1,122 @@
+# How To: Create a DID
+
+## Background
```
andresuribe87, line 3: Would suggest a TL;DR section with a oneliner curl command. WDYT?

doc/howto/did.md
```diff
@@ -0,0 +1,122 @@
...
+
+A [DID (Decentralized Identifier)](https://www.w3.org/TR/did-core/) is a unique identifier that can be used to identify a person, organization, or thing. DIDs are associated with documents called DID Documents. These documents contain
+cryptographic key material, service endpoints, and other useful information.
+
+DIDs are a core component of SSI (Self-Sovereign Identity) systems. DIDs are specified according to "methods" which
+define how the DID is created and how it can be used. A list of existing [DID methods can be found here](https://www.w3.org/TR/did-spec-registries/#did-methods).
+
+Importantly, DIDs in the SSI Service are _fully custodial_. This means the private keys associated with DID Documents are managed by the service and never leave its boundaries. For some DID methods, such as `did:web` it's possible to add multiple keys to the DID document, and it's possible for these additional keys to be added outside the service. This is a more advanced concept that the service may support in the future.
```
andresuribe87, line 11: !! super important.

doc/swagger.yaml
```diff
@@ -2198,12 +2198,12 @@ info:
...
     name: TBD
     url: https://github.com/TBD54566975/ssi-service/issues
-  description: '{{.Desc}}'
+  description: 'The Self Sovereign Identity Service: Managing DIDs, Verifiable Credentials,
+    and more!'
   license:
     name: Apache 2.0
     url: http://www.apache.org/licenses/LICENSE-2.0.html
   title: SSI Service API
-  version: '{{.SVN}}'
```
andresuribe87, line 2206: I think we should PIN this in source code, and remove from the configuration file. What's the purposed of having it be configurable? We have no logic that's making decisions off of it.

pkg/server/server.go
```diff
@@ -72,23 +68,8 @@ func NewSSIServer(shutdown chan os.Signal, cfg config.SSIServiceConfig) (*SSISer
...
-				logrus.WithError(err).Warnf("unable to delete %q during shutdown", tmpFile.Name())
-			}
-			return nil
-		})
-
-		engine.StaticFile("swagger.yaml", tmpFile.Name())
-		engine.GET(SwaggerPrefix, ginswagger.WrapHandler(swaggerfiles.Handler, ginswagger.URL("/swagger.yaml")))
-	}
+	engine.StaticFile("swagger.yaml", "./doc/swagger.yaml")
+	engine.GET(SwaggerPrefix, ginswagger.WrapHandler(swaggerfiles.Handler, ginswagger.URL("/swagger.yaml")))
```
andresuribe87, line 72: Removing this makes it difficult to specify the server urls dynamically. This is important to actually have a nice swagger config.

decentralgabe: yes, but doing it with a template file is not a great solution because it leaves ugly template files to render on the website

using swaggo's default override system as we had before doesn't have this issue, so I will re-add that

doc/swagger.yaml
decentralgabe: agreed

doc/howto/did.md
decentralgabe: added below

doc/swagger.yaml
decentralgabe: we have this issue to track https://github.com/TBD54566975/ssi-service/issues/502 don't want to overload this change set

pkg/server/server.go
andresuribe87: As long as there is a way of handling that use case, then this lgtm.

andresuribe87 (approved):
Approving subject to having a way of specifying server urls in the swagger file.

decentralgabe:
@michaelneale WIP yes, will take the VC stuff in a follow up PR didn't want this to get too big!

PR #625, opened 2023-08-01
config moved to serviceinfo singleton

### LINKED ISSUE #502: Config should not expose non-configurable values

thereisnogabe:
fix #502 

remove config that shouldn't have been configurable; replaced by a service info singleton


config/info.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+var (
+	serviceInfo *ServiceInfo
+	once        sync.Once
+)
+
+// GetServiceInfo provides ServiceInfo as a singleton
+func GetServiceInfo() *ServiceInfo {
+	once.Do(func() {
```
andresuribe87, line 22: I found it surprising to use sync mechanisms only in this function, but not in the underlying object. All the setter methods are not concurrency-safe. In particular, `SetServicePath` may fail if it's called concurrently by multiple go-routines. 

I recommend not using `once`, and keep this simple by exporting `serviceInfo`, as well as all the fields within it.

config/info.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	"sync"
+
+	"github.com/tbd54566975/ssi-service/pkg/service/framework"
+)
+
+const (
+	ServiceName    = "ssi-service"
+	ServiceVersion = "0.0.3"
```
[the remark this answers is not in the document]
decentralgabe: added

config/info.go
decentralgabe: yeah probably better to keep it simple

decentralgabe: read about its usage here https://thedevelopercafe.com/articles/singleton-in-golang-839d8610958b

the use of `once.Do` enforces that only one caller will obtain a lock on the singleton at a time, so the setter methods are indeed thread safe

andresuribe87: That applies only to callers of the GetServiceInfo function. The same does not hold for different threads that already have a handle to the `*ServiceInfo` instance, and both calling `SetServicePath` at the same time.

decentralgabe: It's not a risk outside of the config package, I can move it to its own package if that would remove your concern

decentralgabe: re-refactored to remove concurrency controls; added a warning

pkg/server/router/credential_test.go
```diff
@@ -26,6 +27,9 @@ import (
 
 func TestCredentialRouter(t *testing.T) {
 
+	config.SetAPIBase("http://localhost:1234")
```
andresuribe87, line 30: `apiBase` is a global setting, but there are two tests (maybe more) setting it. Tests in different files run concurrently in golang, so test may fail when they're run because they both modify the same variable.

Having a single static variable that's not injectable, makes it hard to test. Consider passing a reference to the singleton as a parameter to `NewCredentialService`. This way, you can then test using a simple mock.

decentralgabe: I would rather not introduce dependencies to each service on what's supposed to be a global config
as an alternative I'll refactor to have all tests use the same API base

decentralgabe: another option is refactoring the Set methods to SetOnce semantics so they cannot be overwritten

andresuribe87 (approved):

PR #627, opened 2023-08-02
Retrieve pagination parameters from the correct location

### LINKED ISSUE #626: The pageSize and pageToken parameters are perhaps not working when calling the ListSubmissions function.
I found that when calling

http://localhost:3000/v1/presentations/submissions?pageSize=1&filter=status="approved"
the filter parameter is effective, but the pageSize parameter is not.

Furthermore, I found that

`pageSizeStr := framework.GetParam(c, *PageSizeParam*)`

and

`queryPageToken := framework.GetParam(c, *PageTokenParam*)`

in the `ParsePaginationParams` function both return nil.

I modified them to

`pageSizeStr := framework.GetQueryValue(c, *PageSizeParam*)`

and

`queryPageToken := framework.GetQueryValue(c, *PageTokenParam*)`

and it seems to work correctly in terms of the number of results returned. Considering that both functions `ListDIDsByMethod` and `ListSubmissions` that call `ParsePaginationParams` are GET methods, would `GetQueryValue` be a better choice?

andresuribe87:
# Overview
Fixes #626

# Description
Parameters are before the URL path (i.e. before the `?`). Query parameters come after.

# How Has This Been Tested?
Tests were updated to fail before changes were made. And are passing after the changes were made.


decentralgabe (approved):

pkg/server/pagination/pagination.go
```diff
@@ -31,7 +31,7 @@ const (
 // execution is responded to using the passed in gin.Context. The return value corresponds to whether there was an
 // error within the function.
 func ParsePaginationParams(c *gin.Context, pageRequest *PageRequest) bool {
```
decentralgabe, line 33: rename to `ParsePaginationQueryValues`?

andresuribe87: Done in [80a4239](https://github.com/TBD54566975/ssi-service/pull/627/commits/80a42397a33b11e0e6487fe0a2e485b6d3323eac)

PR #646, opened 2023-08-09
Added did ion deactivation endpoint.

andresuribe87:
# Overview
As part of #340, this PR adds support for deactivation. See the API documentation for more details.

# Description
This PR involves a couple of changes so that we return resolution results.

# How Has This Been Tested?
Unit tests.

### DIFF
--- a/doc/swagger.yaml
+++ b/doc/swagger.yaml
@@ -1567,6 +1567,15 @@ definitions:
     - '@context'
     - linked_dids
     type: object
+  pkg_server_router.DeactivateDIDRequest:
+    type: object
+  pkg_server_router.DeactivateDIDResponse:
+    properties:
+      resolutionResult:
+        allOf:
+        - $ref: '#/definitions/resolution.Result'
+        description: The resolution result received from the ION node.
+    type: object
   pkg_server_router.GetApplicationResponse:
     properties:
       application:
@@ -2280,6 +2289,17 @@ definitions:
         description: The `updateCommitment` property in https://identity.foundation/sidetree/spec/#did-resolver-output
         type: string
     type: object
+  resolution.Result:
+    properties:
+      '@context':
+        type: string
+      didDocument:
+        $ref: '#/definitions/did.Document'
+      didDocumentMetadata:
+        $ref: '#/definitions/resolution.DocumentMetadata'
+      didResolutionMetadata:
+        $ref: '#/definitions/resolution.Metadata'
+    type: object
   schema.JSONSchema:
     additionalProperties: {}
     type: object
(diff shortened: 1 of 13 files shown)


doc/swagger.yaml
```diff
@@ -2927,6 +2947,53 @@ paths:
       summary: Updates a DID document.
       tags:
       - DecentralizedIdentityAPI
+  /v1/dids/{method}/{id}/deactivation:
```
decentralgabe, line 2950: this can be more restful as a delete to `/v1/dids/{method}/{id}`

pkg/service/did/handler.go
```diff
@@ -32,6 +32,9 @@ type MethodHandler interface {
 
 	// SoftDeleteDID marks the given DID as deleted. It is not removed from storage.
 	SoftDeleteDID(ctx context.Context, request DeleteDIDRequest) error
+
+	// Resolve returns the resolution result of the given DID according to https://w3c-ccg.github.io/did-resolution/#did-resolution-result.
+	Resolve(ctx context.Context, did string) (*resolution.Result, error)
```
decentralgabe, line 37: we have a separate resolution endpoint...and a get here...do we need both?

pkg/service/did/ion.go
```diff
@@ -658,3 +728,82 @@ func (h *ionHandler) storeUpdateStates(ctx context.Context, tx storage.Tx, id st
...
+		{
+			Namespace: deactivateRequestNamespace,
+			Key:       deactivateRequestKey,
+		},
+	}
+	deactivatedDID, err := h.storage.db.Execute(ctx, func(ctx context.Context, tx storage.Tx) (any, error) {
+		if storedDID.Deactivated {
+			return storedDID, nil
+		}
+		storedDID.Operations = append(storedDID.Operations, deactivateRequest)
```
decentralgabe, line 781: this seems off - shouldn't the result be the processed deactivate request?

pkg/service/did/ion.go
```diff
@@ -658,3 +728,82 @@ func (h *ionHandler) storeUpdateStates(ctx context.Context, tx storage.Tx, id st
...
+			return nil, errors.Wrap(err, "storing DID in storage")
+		}
+		return storedDID, nil
+	}, watchKeys)
+
+	if err != nil {
+		return nil, errors.Wrap(err, "executing deactivation")
+	}
+
+	return &DeactivateIONDIDResponse{Result: deactivatedDID.(*ionStoredDID).DeactivationResponse}, nil
```
decentralgabe, line 808: also it seems as the keystore should revoke related keys

doc/swagger.yaml
decentralgabe: hm, I see. awkward because we both need to handle the service's concept of deletion and the did methods
I'm inclined to merge the two...
Something lke -- if the DID method supports deactivation, the `DELETE` to `/v1/dids/{method}/{id}` does both delete and deactivation. If the DID method does not support deactivation, we just do a delete.

Reason being, it's more confusing to support both separately. What do you think?

pkg/service/did/ion.go
decentralgabe: disregard

doc/swagger.yaml
andresuribe87: I like that, will update.

pkg/service/did/ion.go
andresuribe87: I think we're eventually going to need to communicate to devs that some new state hasn't been published. That said, I agree that it's too complex for now.

I made my best to ensure that all the properties were correct.

decentralgabe (approved):

PR #658, opened 2023-08-15
Added batch update for statuses

### LINKED ISSUE #347: Support for Batch Operations

andresuribe87:
# Overview
This PR adds a batch endpoint for updating statuses of credentials. It is the last PR that fixes #347

# Description
Added the `v1/credentials/status/batch` endpoint. See the swagger spec file for documentation on the request/response schemas.

# How Has This Been Tested?
Unit and integration tests.

### DIFF
--- a/pkg/server/router/credential.go
+++ b/pkg/server/router/credential.go
@@ -339,6 +339,72 @@ type UpdateCredentialStatusResponse struct {
 	Suspended bool `json:"suspended"`
 }
 
+type SingleUpdateCredentialStatusRequest struct {
+	// ID of the credential who's status should be updated.
+	ID string `json:"id"`
+	UpdateCredentialStatusRequest
+}
+
+type BatchUpdateCredentialStatusRequest struct {
+	// Required. The list of update credential requests. Cannot be more than the config value in `services.credentials.batch_update_status_max_items`.
+	Requests []SingleUpdateCredentialStatusRequest `json:"requests" maxItems:"100" validate:"required,dive"`
+}
+
+func (r BatchUpdateCredentialStatusRequest) toServiceRequest() *credential.BatchUpdateCredentialStatusRequest {
+	var req credential.BatchUpdateCredentialStatusRequest
+	for _, routerReq := range r.Requests {
+		serviceReq := routerReq.toServiceRequest(routerReq.ID)
+		req.Requests = append(req.Requests, serviceReq)
+	}
+	return &req
+}
+
+type BatchUpdateCredentialStatusResponse struct {
+	CredentialStatuses []credential.Status `json:"credentialStatuses"`
+}
+
+// BatchUpdateCredentialStatus godoc
+//
+//	@Summary		Batch Update a Verifiable Credential's status
+//	@Description	Updates the status all a batch of Verifiable Credentials.
+//	@Tags			Credentials
+//	@Accept			json
+//	@Produce		json
+//	@Param			request	body		BatchUpdateCredentialStatusRequest	true	"request body"
+//	@Success		201		{object}	BatchUpdateCredentialStatusResponse
+//	@Failure		400		{string}	string	"Bad request"
+//	@Failure		500		{string}	string	"Internal server error"
+//	@Router			/v1/credentials/status/batch [put]
+func (cr CredentialRouter) BatchUpdateCredentialStatus(c *gin.Context) {
+	var batchRequest BatchUpdateCredentialStatusRequest
+	invalidCreateCredentialRequest := "invalid batch update credential request"
+	if err := framework.Decode(c.Request, &batchRequest); err != nil {
+		errMsg := invalidCreateCredentialRequest
+		framework.LoggingRespondErrWithMsg(c, err, errMsg, http.StatusBadRequest)
+		return
+	}
+
+	batchUpdateMaxItems := cr.service.Config().BatchUpdateStatusMaxItems
+	if len(batchRequest.Requests) > batchUpdateMaxItems {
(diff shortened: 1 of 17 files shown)


pkg/server/router/credential.go
```diff
@@ -339,6 +339,72 @@ type UpdateCredentialStatusResponse struct {
 	Suspended bool `json:"suspended"`
 }
 
+type SingleUpdateCredentialStatusRequest struct {
```
decentralgabe, line 342:
```suggestion
type UpdateCredentialStatusRequest struct {
```

pkg/server/router/credential.go
```diff
@@ -339,6 +339,72 @@ type UpdateCredentialStatusResponse struct {
...
+	// ID of the credential who's status should be updated.
+	ID string `json:"id"`
```
decentralgabe, line 344: this should be required

pkg/server/router/credential.go
```diff
@@ -339,6 +339,72 @@ type UpdateCredentialStatusResponse struct {
...
+	Requests []SingleUpdateCredentialStatusRequest `json:"requests" maxItems:"100" validate:"required,dive"`
+}
+
+func (r BatchUpdateCredentialStatusRequest) toServiceRequest() *credential.BatchUpdateCredentialStatusRequest {
+	var req credential.BatchUpdateCredentialStatusRequest
+	for _, routerReq := range r.Requests {
+		serviceReq := routerReq.toServiceRequest(routerReq.ID)
+		req.Requests = append(req.Requests, serviceReq)
+	}
+	return &req
```
decentralgabe, line 359: why return a ptr?

pkg/server/server.go
```diff
@@ -218,6 +218,7 @@ func CredentialAPI(rg *gin.RouterGroup, service svcframework.Service, webhookSer
 	// Credential Status
 	credentialAPI.GET("/:id"+StatusPrefix, credRouter.GetCredentialStatus)
 	credentialAPI.PUT("/:id"+StatusPrefix, credRouter.UpdateCredentialStatus)
+	credentialAPI.PUT(StatusPrefix+"/batch", credRouter.BatchUpdateCredentialStatus)
```
decentralgabe, line 221: change `/batch` to a const which can be used elsewhere

pkg/server/server_credential_test.go
```diff
@@ -31,6 +31,109 @@ func TestCredentialAPI(t *testing.T) {
 
 	for _, test := range testutil.TestDatabases {
 		t.Run(test.Name, func(tt *testing.T) {
+			tt.Run("Batch Update Credential Status", func(ttt *testing.T) {
```
decentralgabe, line 34: we need a few different test cases here

1. batch of 0
2. batch of 1-n (all valid)
3. batch of 1-n (mixed invalid and valid credentials, invalid = cred doesn't exist, cred status already set)

Another test we might need here and elsewhere is performing an update where the key no longer exists (revoked/deleted)

pkg/service/credential/model.go
```diff
@@ -85,8 +85,22 @@ type UpdateCredentialStatusRequest struct {
...
 
 type UpdateCredentialStatusResponse struct {
-	Revoked   bool `json:"revoked" validate:"required"`
-	Suspended bool `json:"suspended" validate:"required"`
+	Status
+}
+
+type Status struct {
+	// ID of the credentials who's status this object represents.
+	ID        string `json:"id,omitempty"`
```
decentralgabe, line 93: not required?

pkg/service/credential/service.go
```diff
@@ -675,3 +658,71 @@ func (s Service) BatchCreateCredentials(ctx context.Context, batchRequest BatchC
...
+	return batchResponse, nil
+}
+
+func (s Service) statusListCredentialWatchKey(ctx context.Context, id string) (*storage.WatchKey, error) {
+	gotCred, err := s.storage.GetCredential(ctx, id)
+	if err != nil {
+		return nil, errors.Wrap(err, "reading credential")
+	}
+
+	if gotCred.Credential.CredentialStatus == nil {
```
decentralgabe, line 708: I would recommend a new helper like `HasCredentialStatus` which can also check that `gotCred` is not nil and `gotCred.Credential` is not nil, or else we're at risk for NPEs

pkg/service/credential/service.go
```diff
@@ -675,3 +658,71 @@ func (s Service) BatchCreateCredentials(ctx context.Context, batchRequest BatchC
...
+		return nil, sdkutil.LoggingNewErrorf("credential %q has no credentialStatus field", gotCred.LocalCredentialID)
+	}
+
+	statusPurpose := gotCred.Credential.CredentialStatus.(map[string]any)["statusPurpose"].(string)
```
decentralgabe, line 712: `GetStatusPurpose` method would be helpful

decentralgabe (changes requested):
see comments

pkg/server/server.go
andresuribe87: Done

pkg/server/server_credential_test.go
andresuribe87: Added tests for all batch behaviors.

> Another test we might need here and elsewhere is performing an update where the key no longer exists (revoked/deleted)

I'm unsure about some of the details of the behavior you're trying to test. Mind creating an issue and we can discuss?

pkg/service/credential/model.go
andresuribe87: No, the usage in `UpdateCredentialStatusResponse` is optional.

decentralgabe (approved):

Issue #640, opened 2023-08-04
Allow fully configurable URIs for Status and Credentials


decentralgabe:
At present, using the `api_host` we can set a per-service value to be stamped in all VCs and Status Lists the service produces. We can take this a step further, by allowing per-service configuration of this value...and not just for the base host, but for the entire path.

- An implementer should be able to set any path to externally access VCs/Status Lists.
- An implementer may choose to not have credentials be externally resolvable (this should be the default - for privacy reasons)
- An implementer may choose to have a different URI for hosting a status list separate to the service itself

michaelneale:
@decentralgabe is this a #goodfirstissue type of thing for contributors potentially?

michaelneale:
relevant line of code: https://github.com/TBD54566975/ssi-service/blob/main/pkg/service/credential/status.go#L72 

```go
statusListURI := fmt.Sprintf("%s/status/%s", config.GetServicePath(framework.Credential), statusListID)
```

(just so I remember it)

michaelneale:
@decentralgabe so am I right in thinking if fully specified statusListURI should be a constant url for the whole status list?

michaelneale:
I think a common case would be a service in front of ssi which would choose to expose the statuslist (and only the status list). Could the api_host be used for this or a separate config?

decentralgabe:
> @decentralgabe so am I right in thinking if fully specified statusListURI should be a constant url for the whole status list?

yes though I imagine what we want is the ability to set a base path and then let the service handle the ID generation

so you would configure `statusBaseURL` to `https://my-website.com/status` and then the service would add `https://my-website.com/status/{status-list-id}`

michaelneale:
yep I think you are right - given that is the preferred pattern, and makes it clearer what they are for. statusBaseURL could perhaps be optionally a full path in which case it won't append the ID (but that means it is up to implementers to ensure status less is updated elsewhere, so maybe that is a bit premature and YAGNI for now)

decentralgabe:
how would it work without the ID given that there are many status lists? for each <schema, issuer> pair?

decentralgabe:
fixed in #660

michaelneale:
thanks @decentralgabe

PR #660, opened 2023-08-17
adding in seperate status base URI

michaelneale:
addressing: https://github.com/TBD54566975/ssi-service/issues/640

@decentralgabe am I on the right track?

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -68,6 +68,7 @@ type ServicesConfig struct {
 	StorageProvider string           `toml:"storage" conf:"default:bolt"`
 	StorageOptions  []storage.Option `toml:"storage_option"`
 	ServiceEndpoint string           `toml:"service_endpoint" conf:"default:http://localhost:8080"`
+	StatusEndpoint  string           `toml:"status_endpoint" conf:"default:http://localhost:8080"`
 
 	// Application level encryption configuration. Defines how values are encrypted before they are stored in the
 	// configured KV store.
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -65,6 +65,7 @@ func NewSSIServer(shutdown chan os.Signal, cfg config.SSIServiceConfig) (*SSISer
 
 	// make sure to set the api base in our service info
 	config.SetAPIBase(cfg.Services.ServiceEndpoint)
+	config.SetStatusBase(cfg.Services.StatusEndpoint)
 
 	// service-level routers
 	engine.GET(HealthPrefix, router.Health)
--- a/pkg/service/credential/status.go
+++ b/pkg/service/credential/status.go
@@ -67,10 +67,16 @@ func (s Service) createStatusListEntryForCredential(ctx context.Context, credID 
 	}, nil
 }
 
+func getStatusURI(statusBase string, statusListID string) string {
+	if len(statusBase) > 0 {
+		return fmt.Sprintf("%s/status/%s", statusBase, statusListID)
+	}
+	return fmt.Sprintf("%s/status/%s", config.GetServicePath(framework.Credential), statusListID)
+}
+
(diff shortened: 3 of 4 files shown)


pkg/service/credential/status.go
```diff
@@ -67,10 +67,16 @@ func (s Service) createStatusListEntryForCredential(ctx context.Context, credID
 	}, nil
 }
 
+func getStatusURI(statusBase string, statusListID string) string {
```
decentralgabe, line 70: this method isn't needed since you'll be setting the base with `SetStatusBase`

config/config.go
```diff
@@ -68,6 +68,7 @@ type ServicesConfig struct {
 	StorageProvider string           `toml:"storage" conf:"default:bolt"`
 	StorageOptions  []storage.Option `toml:"storage_option"`
 	ServiceEndpoint string           `toml:"service_endpoint" conf:"default:http://localhost:8080"`
+	StatusEndpoint  string           `toml:"status_endpoint" conf:"default:http://localhost:8080"`
```
decentralgabe, line 71: since this has a default your check blow of len > 0 will never fail

pkg/service/credential/status.go
```diff
@@ -67,10 +67,16 @@ func (s Service) createStatusListEntryForCredential(ctx context.Context, credID
...
+	if len(statusBase) > 0 {
+		return fmt.Sprintf("%s/status/%s", statusBase, statusListID)
```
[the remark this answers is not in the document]
decentralgabe: an alternate approach is that the base includes a path, so this just becomes

```go
		return fmt.Sprintf("%s/%s", statusBase, statusListID)
```

decentralgabe (commented):
see comments

pkg/service/credential/status.go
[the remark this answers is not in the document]
michaelneale: yeah - done.

config/config.go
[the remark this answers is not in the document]
michaelneale: yeah changed it to be empty

michaelneale:
@decentralgabe yeah made those changes, sorry didn't get to add tests specifically for this.

decentralgabe (approved):
if you're able to test locally and verify that would be great - then feel free to merge

integration/common.go
```diff
@@ -51,6 +51,7 @@ func init() {
 	config.SetAPIBase(endpoint)
 	config.SetServicePath(framework.Credential, "/credentials")
 	config.SetServicePath(framework.Schema, "/schemas")
+	config.SetStatusBase(fmt.Sprintf("%s/status", config.GetServicePath(framework.Credential)))
 }
```
decentralgabe, line 55:
```suggestion
	config.SetServicePath(framework.Schema, "/schemas")
}
```

integration/common.go
```diff
@@ -20,17 +20,15 @@ import (
...
 	"github.com/tbd54566975/ssi-service/internal/keyaccess"
 	"github.com/tbd54566975/ssi-service/internal/util"
 	"github.com/tbd54566975/ssi-service/pkg/server/router"
-	"github.com/tbd54566975/ssi-service/pkg/service/framework"
 )
 
 const (
 	// Note: for local testing change this to port 3000
-	endpoint       = "http://localhost:8080/"
+	endpoint       = "http://localhost:3000/"
```
decentralgabe, line 31:
```suggestion
	endpoint       = "http://localhost:8080/"
```

pkg/server/server.go
```diff
@@ -208,6 +209,13 @@ func CredentialAPI(rg *gin.RouterGroup, service svcframework.Service, webhookSer
 	// make sure the credential service is configured to use the correct path
 	config.SetServicePath(svcframework.Credential, CredentialsPrefix)
 
+	// allows for a custom URI to be used for status list credentials, if not set, we use the default path
+	if statusEndpoint != "" {
```
decentralgabe, line 213: this needs to be initialized after the cred prefix is set

PR #672, opened 2023-08-30
Refactoring so error causality is easier to follow

### LINKED ISSUE #635: Remove the word "error" (and other negatives) from `error.Wrap` messages
The fact that it's an error is already or that an operation couldn't be performed is already implicit within the fact that the type of the returned value is `error`. As such, things like `error.Wrap(err, "error calculating X")` should be replaced by `error.Wrap(err, "calculating X")`. The details of the root cause of the failure should come from the `err` variable.

andresuribe87:
# Overview
Fixes #635


andresuribe87:
@nitro-neal or @decentralgabe could I get a stamp here plz?

decentralgabe (approved):

